PR #115, opened 2021-12-02
Add Logout button to user page

brekk:
Dec-02-2021 13-51-28

### DIFF
--- a/components/Navbar/LoginButton.tsx
+++ b/components/Navbar/LoginButton.tsx
@@ -50,7 +50,7 @@ type StaticProps = {
 }
 
 const StaticButton = ({ href }: StaticProps) => (
-  <Link href={href}>
+  <Link href={href} passHref>
     <a>
       <span>
         Login<span className="md:hidden"> to Testnet</span>
@@ -102,6 +102,8 @@ const UserButton = ({ id, graffiti, visible }: ApiUserMetadataUI) => (
       <a
         className={clsx(
           styles.arrowed,
+          'pointer-events-none',
+          'md:pointer-events-auto',
           'md:hidden',
           'md:opacity-0',
           { 'md:opacity-100': visible },
--- a/components/Navbar/LoginButton.tsx
+++ b/components/Navbar/LoginButton.tsx
@@ -1,10 +1,13 @@
 import { useState, ReactNode, MouseEventHandler } from 'react'
 import Link from 'next/link'
 import clsx from 'clsx'
-
-import { ApiUserMetadata } from 'apiClient/types'
 import { LoginContext } from 'contexts/LoginContext'
+
+// TODO: disable this before merging!
+import useQuery from 'hooks/useQuery'
+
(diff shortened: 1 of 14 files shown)


dguenther:
I think having the logout feature is more useful than not, so wouldn't consider these blockers, but a few things I ran into when testing:
- Hovering over the Sign Out text in mobile/narrow view sets the username pill to the hovered state.
- Similar to logging in (which we have a ticket for), the menu pill doesn't update to remove your username when you log out. Since there's also not a "You have been logged out" toast, it looks like you get redirected to the login screen but are still logged in. I don't think both are needed, but without either there's potential for confusion.

dguenther (approved):

pages/logout.tsx
```diff
@@ -2,13 +2,18 @@ import { useEffect } from 'react'
...
 export default function Logout() {
   useEffect(() => {
     const sayGoodbye = async () => {
       if (!magic) return false
       await magic.user.logout()
-      Router.push((await magic.user.isLoggedIn()) ? '/leaderboard' : '/login')
+      Router.push(
+        (await magic.user.isLoggedIn())
+          ? `/leaderboard`
+          : `/login?toast=${btoa('You have been logged out.')}`
```
brekk, line 15: > Similar to logging in (which we have a ticket for), the menu pill doesn't update to remove your username when you log out. Since there's also not a "You have been logged out" toast, it looks like you get redirected to the login screen but are still logged in. I don't think both are needed, but without either there's potential for confusion.

@dguenther How does this work for you?

components/Navbar/LoginButton.tsx
```diff
@@ -1,68 +1,212 @@
...
+const UserButton = ({ id, graffiti, visible }: ApiUserMetadataUI) => (
+  <div
+    className={clsx(
+      styles.userButton,
+      'absolute',
+      'flex',
+      'flex-col',
+      'top-[0.75rem]',
+      'pointer-events-none',
+      'md:pointer-events-auto',
```
brekk, line 80: > Hovering over the Sign Out text in mobile/narrow view sets the username pill to the hovered state.

@dguenther So this one is a little tricky, basically I eat the hover events on mobile / base styling and then swap back to regular at the next breakpoint. However, on desktop at mobile breakpoints, because of the nature of `pointer-events` I wasn't able to make it so that the mouse looks right.

pointerless
 
I think that's about the best I can do without investing a bunch more time and it's a smaller subset of the original problem b/c on mobile you won't have any hover events either way.

pages/logout.tsx
dguenther: That works 👍

components/Navbar/LoginButton.tsx
dguenther: Sounds good, that change seems like an improvement to me, I'm okay punting on this for now.

dguenther (approved):

PR #122, opened 2021-12-03
Hide settings tab on non-authed user page

### LINKED ISSUE #119: collusion graffiti
I have tried sign in but not sign in site. Later, I have registered with other e-mail and diffirent graffiti. They are collusion. the pages is below..
Screenshot_1
Screenshot_2

dguenther:
The settings tab currently shows up on every user page when you're logged in, which led to some confusion around whether you're viewing settings for that user or for the logged-in user. This change makes it only appear when the authed user ID matches the user page's user ID.

I renamed a bunch of props here to what I thought was more clear, I'm fine rolling those renames back if it doesn't seem like an improvement 👍 

Fixes #119

### DIFF
--- a/components/user/Tabs/SettingsContent.tsx
+++ b/components/user/Tabs/SettingsContent.tsx
@@ -7,40 +7,40 @@ import { useField } from 'hooks/useForm'
 import { FIELDS } from 'pages/signup'
 
 type Props = {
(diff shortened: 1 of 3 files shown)


brekk (approved):
Yes, this change makes good sense to me. 👍

PR #126, opened 2021-12-06
IRO-1307 - Address UI lag in rendering Magic Link state

brekk:
- Remove LoginContext usage
- Number explicit error boundaries
- thread props for loginContext, coming from `pages/_app` directly. This ends up being both more reliable and ultimately more readable / easier to reason about.
- refactor to use `useRouter` instead of `Router.push`


hooks/useProtectedRoute.ts
```diff
@@ -14,23 +14,25 @@ export function useProtectedRoute({
   redirect = '',
   timeout = -1,
 }: ProtectedProps) {
-  const { checkLoggedIn, status, metadata, magicMetadata } = useLogin({
+  const $router = useRouter()
+  const loginState = useLogin({
```
dguenther, line 18: This is part of the issue -- useProtectedRoute is establishing its own login state store by calling `useLogin` instead of using the LoginContext. If clients are reading from it, they'll have a different view of the login state.

hooks/useLogin.ts
```diff
@@ -36,7 +43,9 @@ export function useLogin(config: LoginProps = {}) {
       try {
```
dguenther, line 43: This isn't causing any user-perceivable issues, but we're making too many API calls because this function fires every time `$status` changes. On master it makes one call initially when status is `loading`, one call when status is `loaded`, and potentially an additional call when status is `forced` if the timeout happens.

pages/logout.tsx
```diff
@@ -1,21 +1,22 @@
...
+import { useRouter } from 'next/router'
 import Loader from 'components/Loader'
 import { encode as btoa } from 'base-64'
 
 export default function Logout() {
+  const $router = useRouter()
   useEffect(() => {
     const sayGoodbye = async () => {
       if (!magic) return false
       await magic.user.logout()
```
dguenther, line 12: The LoginContext/useLogin needs to know that the login state may have changed after this call. If `magic` exposed an event for when login state changes, we could use that in LoginContext/useLogin, but since it doesn't, we should either centralize all auth logic in useLogin and export functions to be used elsewhere, or export functions that allow for modifying the state of useLogin or retriggering `checkLoggedIn`.

pages/_app.tsx
```diff
@@ -1,12 +1,12 @@
...
 import type { AppProps } from 'next/app'
-import { LoginContext } from 'contexts/LoginContext'
-import useLogin from 'hooks/useLogin'
 import ResponsiveToolkit from 'components/ResponsiveToolkit'
+import { useLogin } from 'hooks/useLogin'
 
-function MyApp({ Component, pageProps }: AppProps) {
-  const $loginValue = useLogin({ timeout: 1500 })
+function MyApp({ Component: Page, pageProps }: AppProps) {
+  const $login = useLogin()
```
dguenther, line 8: This loading timeout is okay to disable now? Seems like disabling it would fix the case where you'd skip the loading screen after the timeout and see the page in an unauthed state, but potentially lock the site behind the loading screen if the API goes down

pages/login.tsx
```diff
@@ -115,6 +120,21 @@ export default function Login() {
...
+          checkLoggedIn: () => false,
+          checkLoading: () => false,
+          checkFailed: () => false,
+          status: STATUS.LOGGED_OUT,
+          setStatus: () => null,
+          error: null,
+          magicMetadata: null,
+          metadata: null,
+        }
+      : loginContext
```
dguenther, line 135: This is a hack -- we should update the source of the data to a logged-out state when the logout happens rather than covering up the data and switching on the contents of the toast.

dguenther (changes requested):
This doesn't work as expected. When logging out, after the redirect happens, the Login page appear to be logged out, but if you click on Leaderboard, the navbar appears to be logged in again.

The source of the issue is the multiple instances of `useLogin` created by `useProtectedRoute`. The site only needs one source of truth for whether the user is logged in or not, so we can centralize that in the single `useLogin` in `_app.tsx`. `useLogin` should either contain all of the auth logic and export functions to control it, or export functions to control its state so that e.g. logout can clear the state and the login callback can update or trigger a reload of the state.

The use of Context is irrelevant to fixing this issue, but I'm not opposed to prop-drilling since there aren't too many components in the component tree anyway.

brekk:
> The source of the issue is the multiple instances of `useLogin` created by `useProtectedRoute`. 

Yeah, you're right. I'm not sure when I made that mistake but that's correct. Will try to address this morning.

pages/_app.tsx
brekk: That was my concern; I guess we have to weigh the difference between "it looks ok but maybe doesn't work" and "it never loads so I can't use it (and maybe it's also broken)". I tentatively think the latter is preferable but not certain on that.

hooks/useLogin.ts
brekk: I added a condition here: https://github.com/iron-fish/website-testnet/pull/126/files#diff-787a09031878f88df5cd5bccb34bded89f4643ec5a089b98a9cfbff465e93019R46 which should prevent successive re-renders

pages/_app.tsx
brekk: I chose the latter for now, we can revisit if it's untenable.

pages/logout.tsx
brekk: I ended up adding these two state resets: https://github.com/iron-fish/website-testnet/pull/126/files#diff-ccc3bc2a49ec13caa0caab60e351ddd83198bd11352f1d051b7c70876e284950R22-R23

pages/login.tsx
brekk: I ended up adding these two state resets: https://github.com/iron-fish/website-testnet/pull/126/files#diff-ccc3bc2a49ec13caa0caab60e351ddd83198bd11352f1d051b7c70876e284950R22-R23 to replace this change.

hooks/useLogin.ts
```diff
@@ -18,27 +24,30 @@ export enum STATUS {
...
+// const $metadata = useLogin({redirect: '/go-somewhere-if-it-does-not-work'})
 
 export interface LoginProps {
   redirect?: string
-  timeout?: number
 }
+
 export function useLogin(config: LoginProps = {}) {
-  const { redirect, timeout = -1 } = config
+  const $router = useRouter()
```
dguenther, line 34: This use of `useRouter` is triggering additional /me data fetches and executions of the `checkLoggedIn` useEffect when the router is added to the useEffect's dependencies, so we should switch back to Router here.

Next.js recommends using the global Router in useEffects and event handlers, and the useRouter/withRouter in components: https://github.com/vercel/next.js/discussions/18522#discussioncomment-836362

Another related issue https://github.com/vercel/next.js/issues/18127

pages/login.tsx
```diff
@@ -98,12 +103,12 @@ export default function Login() {
         $setLoaded(true)
         $setMessage('Logged in!')
```
dguenther, line 104: There should be a data refetch triggered on the `useLogin` context here.

The login button is currently updating post-redirect due to Next.js useRouter triggering a data refetch as mentioned above, but the user button should update (or start updating) as soon as we know the user has successfully logged in.

dguenther (changes requested):

pages/logout.tsx
```diff
@@ -1,21 +1,38 @@
...
-      if (!magic) return false
-      await magic.user.logout()
-      Router.push(
-        (await magic.user.isLoggedIn())
-          ? `/leaderboard`
-          : `/login?toast=${btoa('You have been logged out.')}`
-      )
+      // eslint-disable-next-line no-console
+      console.log('logging out')
+      if (magic && magic.user && magic.user.logout) {
```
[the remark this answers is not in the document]
brekk: > If magic is defined, magic.user and magic.user.logout will always be defined according to the TS types so it's not necessary to check them.
Agreed, but I wrote this because TS wouldn't stop fighting me on it. Happy to try to revisit but I remember this specifically being the case.

pages/_app.tsx
```diff
@@ -1,12 +1,12 @@
...
 import Head from 'next/head'
 import type { AppProps } from 'next/app'
-import { LoginContext } from 'contexts/LoginContext'
-import useLogin from 'hooks/useLogin'
 import ResponsiveToolkit from 'components/ResponsiveToolkit'
+import { useLogin } from 'hooks/useLogin'
 
-function MyApp({ Component, pageProps }: AppProps) {
-  const $loginValue = useLogin({ timeout: 1500 })
+function MyApp({ Component: Page, pageProps }: AppProps) {
```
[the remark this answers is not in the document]
brekk: I'm amenable to restoring it, but the intention was to increase cohesion between the `/pages` folder and these Page components.

pages/login.tsx
brekk: I guess so -- this is a weird one to me because of how Magic Link works (i.e. This is the original `/login` page that sits around while the user goes to their email and then there's another page with a separate state later.) I kinda considered this original page not really a use-case but I'll add the data here.

hooks/useLogin.ts
brekk: I tried changing to the static `Router.push` but it reverted all of core intent of this PR. (That is, when swapped, the page doesn't update on hitting `/leaderboard` after being redirected from `/login`.) So I am intending to keep this change.

hooks/useLogin.ts
```diff
@@ -18,27 +24,30 @@ export enum STATUS {
...
   const [$metadata, $setMetadata] = useState<ApiUserMetadata | null>(null)
+
   useEffect(() => {
     const checkLoggedIn = async () => {
       try {
-        // this is likely a case where we're working in not-the-browser
-        if ($metadata || !magic || !magic.user) {
-          Promise.reject(new LocalError('Magic instance not available!', 500))
+        // if we're already loaded, quit
+        if ($status === STATUS.LOADED) return
```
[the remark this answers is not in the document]
dguenther: I don't think so -- This `useEffect` runs every time one of its dependencies is updated, and it calls `checkLoggedIn` every time it re-runs. `$status` is one of its dependencies, so for example, in the case where a user does not have a valid login token:

1. `checkLoggedIn` will execute on page load
1. `token` will not be set
1. Control flow proceeds down the `!token` path and calls `$setStatus(STATUS.NOT_FOUND)`
1. Updating the status then triggers the `useEffect` to run again, which will call `checkLoggedIn` again
1. `checkLoggedIn` will make another call to Magic that proceeds down the same control flow as the prior useEffect execution
1. Since the status is set to its now-current value, NOT_FOUND, the useEffect does not run again.

We're still making unnecessary network requests as a result of our state + effect design, which is why the `$status` + `useEffect` design needs to be considered more, but I think it's fine for now since I haven't seen a situation where it has a user-visible effect.

PR #129, opened 2021-12-08
Editable Settings

brekk:
Changes:

- [x] - 🧹 Removed `API_KEY` usage. It's a hold-over from a former age, and now it is dead.
- [x] - 🏋🏽 Added a new function in `apiClient` for updating a user. It's called (surprisingly) `updateUser`. It takes an object blob and passes it to the API via `PUT`.
- [x] - 🔧 Fixed a several bugs in `useField` which would cause it to infinitely re-render
- [x] - Field updates:
  - Disabled email
  - Graffiti with warning-on-touch
  - Split out Discord + Telegram into separate non-required fields.
  - Country Selector with no default
- [x] Wrote a bunch of smoke tests - but to aid readability, I have moved these changes to https://github.com/iron-fish/website-testnet/pull/164
- [x] - 🦾 Improved the behavior of `hooks/useLogin` so that it can be reloaded
- [x] - Updated `/users/[id]` to support picking tabs via query string: `?tab=settings`

Testing Process:

1. Make sure you can log in (use the leaderboard to verify whether you've already created a user on staging)
2. Go to your user (click your username in the upper right). All of the tabs should still work, but now the settings tab should allow one to actually change values. (If you take a look at the network tab you should see the data verifies this) Change some values for fun / profit!
3. You shouldn't be able to see other people's settings


brekk:
This PR will eventually also fix https://github.com/iron-fish/website-testnet/issues/131 (pending some changes on the backend)

hooks/useForm.ts
```diff
@@ -47,70 +52,112 @@ export interface Field extends ProvidedField {
...
   const banSpaces = whitespace === WHITESPACE.BANNED
   const trimSpaces = banSpaces || whitespace === WHITESPACE.TRIMMED
   const [$choice, $setChoice] = useState<string>(
-    provided.isRadioed && radioOption ? radioOption : ''
+    isRadioed && radioOption ? radioOption : ''
   )
   const [$disabled, $setDisabled] = useState<boolean>(false)
-  const [$valid, $setValid] = useState<boolean>(false)
-  const [$touched, $setTouched] = useState<boolean>(false)
+  const [, $setValid] = useState<boolean>(false)
```
dguenther, line 98: I don't think this is necessary at all if it's derivative of the other state, and we're not using the state value.

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+  country: { ...FIELDS.country, useDefault: false },
+  discord: {
+    ...FIELDS.social,
+    required: false,
+    options: undefined,
+    id: 'discord',
+    label: 'Discord',
+    placeholder: 'Your Discord username',
+    isRadioed: false,
+    validation: () => true,
```
dguenther, line 42: This doesn't quite match up with the one-or-the-other requirement from signup, are we getting rid of that for now?

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+  const [$userData, $setUserData] = useState<API.ApiUserMetadata | null>(
+    authedUser
+  )
+
+  const {
+    email: _email = UNSET,
+    graffiti: _graffiti = UNSET,
+    discord: _discord = UNSET,
+    telegram: _telegram = UNSET,
+    country_code: _cc = UNSET,
```
dguenther, line 74: This name can be expanded out to `_country_code` to match the style of the other variable names

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+import { UNSET } from 'utils/forms'
+import { useQueriedToast } from 'hooks/useToast'
+
+import * as API from 'apiClient'
+import { TabType } from './index'
+
 type Props = {
-  authedUser: API.ApiUserMetadata
+  anyBlocksMined: boolean
+  user: API.ApiUser | null
```
dguenther, line 23: Same comment re: the users prop here

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+      telegram,
+      discord,
+      country_code: country,
+    }
+
+    const result = await API.updateUser(authedUser.id, updates)
+
+    if ('error' in result) {
+      const error = '' + result.message
+      $setError(error)
```
dguenther, line 145: Want a `scrollUp` for this case too? It looks like it's used in most other cases when modifying the error or toast.

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+    } else {
+      toast.setMessage('User settings updated')
+      toast.show()
+      $graffiti.setTouched(false)
```
dguenther, line 149: Is this something to do with the validation?

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
 
+    const updates = {
+      email,
+      graffiti,
+      telegram,
+      discord,
+      country_code: country,
+    }
+
+    const result = await API.updateUser(authedUser.id, updates)
```
dguenther, line 141: There should probably be some load or disable state while this is processing, to make it clearer that we're crunching numbers.

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+      !$country?.valid
+    if (invalid) {
+      $setError('Please correct the invalid fields below')
+      scrollUp()
+    } else {
+      $setError(UNSET)
+    }
+    return invalid
+  }, [$email, $graffiti, $telegram, $discord, $country])
+  const update = useCallback(async () => {
```
dguenther, line 115: These effects/callbacks should have line breaks between them so they're a bit easier to distinguish from each other.

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+    const canSee =
+      authedUser &&
+      user &&
+      user.graffiti === authedUser.graffiti &&
+      user.id === authedUser.id
+    // eslint-disable-next-line no-console
+    if (!canSee) {
+      // if you try to go to /users/x/settings but you're not user x
+      onTabChange('weekly')
+      toast.setMessage('You are not authorized to go there')
```
dguenther, line 178: When I change graffiti and hit save, the page says "You are not authorized to go there" and redirects me to the weekly stats page. It also doesn't update the user page and tab title with my new graffiti

dguenther (changes requested):
There are a few changes that aren't necessarily related to Editable Settings that would be great to pull out into separate PRs in the future to make these sort of changes easier to review. I think they're okay for now though, we should fix the main usability bugs with editing graffiti + loading data multiple times and then this should be good to go!

components/Navbar/LoginButton.tsx
```diff
@@ -151,7 +151,7 @@ export const LoginButton = ({ loginContext }: LoginProps) => {
   const on = set(true)
   const off = set(false)
   const { checkLoggedIn, metadata } = loginContext
-  const isLoggedIn = checkLoggedIn()
+  const isLoaded = checkLoggedIn()
```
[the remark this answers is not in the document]
brekk: Agreed, yeah, I actually did the full change in a separate PR but after rebasing this was the result. Happy to address later

components/user/Tabs/SettingsContent.tsx
brekk: Ah, yeah. Discussion in Slack starting around [here](https://iron-fish.slack.com/archives/C02PXM6EUDB/p1640037494111500)

hooks/useForm.ts
brekk: Well we're using the `valid` value downstream from here, just not in the `useEffect` block. https://github.com/iron-fish/website-testnet/pull/129/files#diff-731dce45833d5338e722687a46fcec639ee2c63c0cb87a544aeb7a043d3f0df6R115 I can take a look at removing it but I believe I want both the dynamic `valid` and the ability to programmatically invalidate things downstream.

hooks/useLocalLogin.ts
```diff
@@ -1,26 +1,31 @@
 import { LoginContext, STATUS } from './useLogin'
 
+export const METADATA = {
```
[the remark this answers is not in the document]
brekk: This is used primarily in the testing branch, where it is used directly, but I agree it looks weird in this PR.

package-scripts.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
     dx: concurrent.nps('build', 'lint', 'meta.dependencies'),
     meta: {
       dependencies: {
-        build: `depcruise -c .dependency-cruiser.js -T dot components pages apiClient contexts data definitions hooks public styles utils --progress -x node_modules definitions | dot -T svg > dependency-graph.svg`,
+        build: `depcruise -c .dependency-cruiser.js -T dot components pages apiClient data definitions hooks public styles utils --progress -x node_modules definitions | dot -T svg > dependency-graph.svg`,
```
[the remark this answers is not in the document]
brekk: Yeah, it was yelling because we removed the `contexts` folder. I'll try to keep things separate in the future, but I think this was the outcome of the last task of last year becoming a bigger thing.

package.json
```diff
@@ -7,16 +7,17 @@
     "build": "nps build",
     "start": "nps start",
     "lint": "nps lint",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "ci": "nps test.ci"
```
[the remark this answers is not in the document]
brekk: Good catch. Yes, you're correct.

types.d.ts
```diff
@@ -0,0 +1,10 @@
+import '@testing-library/jest-dom/extend-expect'
+import 'jest-dom/extend-expect'
+import 'jest-fetch-mock'
+
+declare module '*module.css' {
```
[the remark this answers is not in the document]
brekk: Correct, I can slice this one out.

components/user/Tabs/SettingsContent.tsx
dguenther: Hmm, I'm not sure if this totally addresses my question. On the signup page, users are required to either enter a Telegram or enter a Discord username, and they can't leave the field blank (at least one is required). On this settings page, users can delete both fields (neither is required).

One of these changes needs to be made for consistency:
1. This page should validate that one or more of the Telegram or Discord fields has a non-empty value. 
2. OR the signup page should not require a value for Telegram/Discord.

Additionally, our API should not allow empty string/whitespace values to satisfy the field requirement, but I think a frontend change on its own would be good enough for usability.

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+      email,
+      graffiti,
+      telegram,
+      discord,
+      country_code: country,
+    }
+
+    const result = await API.updateUser(authedUser.id, updates)
+
+    if ('error' in result) {
```
dguenther, line 143: This isn't catching the `"Current graffiti '' has already mined blocks"` case -- it shows "User settings updated" instead.

I think the root cause is our API not returning consistent types for error responses, but I imagine that'll be a fair bit of work to get in order, so it'll probably be easier to adjust the frontend to account for it for now.

components/user/Tabs/SettingsContent.tsx
brekk: Fair, I will go the second route here.

components/user/Tabs/SettingsContent.tsx
brekk: Ah yeah, this is to make sure that the graffiti field doesn't pop an error. Will add a comment to that effect.

components/user/Tabs/SettingsContent.tsx
brekk: Added

components/user/Tabs/SettingsContent.tsx
brekk: Added

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,237 @@
...
+    isRadioed: false,
+    validation: () => true,
+  },
+  telegram: {
+    ...FIELDS.social,
+    required: false,
+    options: undefined,
+    id: 'telegram',
+    label: 'Telegram',
+    placeholder: 'Your Telegram username',
```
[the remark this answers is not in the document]
brekk: Addressed, after refactoring `useField`, it wasn't passing this value downstream.

hooks/useForm.ts
```diff
@@ -20,14 +20,19 @@ export interface NameValue {
 }
 
 export interface ProvidedField {
-  id: string
-  label: string
+  defaultErrorText?: string
+  useDefault?: boolean
```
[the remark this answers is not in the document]
brekk: A refactoring victim, now fixed

pages/logout.tsx
```diff
@@ -13,7 +13,7 @@ type LogoutProps = {
 export default function Logout({ loginContext }: LogoutProps) {
   useEffect(() => {
     const sayGoodbye = async () => {
-      const loggedOut = await magic?.user.logout()
+      const loggedOut = await magic?.user?.logout()
```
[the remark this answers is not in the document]
dguenther: Pinging this comment for a follow-up to see if it's resolvable.

hooks/useForm.ts
dguenther: I think the direct test here would be whether removing `$setValid` breaks anything. As it stands, the `valid` that gets put into `$setField` is computed from `$touched` and `$value` and isn't React state, so anything that updates either $touched or $value should cause valid to get recomputed and passed through to `$setField`.

hooks/usePaginatedEvents.ts
```diff
@@ -13,10 +13,10 @@ export function usePaginatedEvents(
   fetchNext: () => void
 } {
   const $hasPrevious = useMemo(
-    () => events?.metadata.has_previous ?? false,
+    () => events?.metadata?.has_previous ?? false,
     [events]
   )
-  const $hasNext = useMemo(() => events?.metadata.has_next ?? false, [events])
+  const $hasNext = useMemo(() => events?.metadata?.has_next ?? false, [events])
```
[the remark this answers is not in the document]
dguenther: Let's not use this operator unless TypeScript requires us to. If there's code in the API that can return metadata as undefined, I'd like to take a look at that, and we can update the TS types for `API.ListEventsResponse`. Otherwise, I'd definitely consider it an error in the frontend if it's somehow undefined in practice, and we should be surfacing the error rather than coalescing it to false.

pages/users/[...details].tsx
```diff
@@ -42,99 +72,122 @@ function displayEventType(type: API.EventType): string {
...
 
   useEffect(() => {
-    let isCanceled = false
+    let isCancelled = false
 
     const fetchData = async () => {
-      if (!router.isReady) {
-        return
-      }
+      try {
```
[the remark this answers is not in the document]
Comment (dguenther) [reply pages/users/[...details].tsx]: Pinging this comment for a follow-up.

hooks/useLogin.ts
```diff
@@ -39,78 +32,79 @@ export function useLogin(config: LoginProps = {}) {
...
-          $setError(new LocalError('No token available.', NO_MAGIC_TOKEN))
-          return
-        }
-        const [magicMd, details] = await Promise.all([
-          magic.user.getMetadata(),
-          getUserDetails(token),
-        ])
+  const reloadUser = useCallback(async () => {
+    try {
+      if (!magic || !magic.user) {
```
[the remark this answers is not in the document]
dguenther: Pinging this comment for a follow-up.

components/user/Tabs/SettingsContent.tsx
dguenther: Not sure if this one is resolved, it looks like the `scrollUp` is still only in the `else`, unless I missed the change.

components/user/Tabs/index.tsx
```diff
@@ -26,7 +29,12 @@ export default function Tabs({
   authedUser,
   activeTab,
   onTabChange,
+  toast,
+  reloadUser,
 }: TabsProps) {
+  const allTimeBlocksMined = allTimeMetrics?.metrics?.blocks_mined?.points ?? 0
```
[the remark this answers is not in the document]
dguenther: Yeah, we should move this over to the testing branch and discuss it there. We should be able to prove that all of these fields are potentially undefined by looking at lines in the API code (and updating the TS types on the frontend), or this is indeed an error that should be surfaced somewhere.

components/user/Tabs/SettingsContent.tsx
dguenther: Re-raising this issue, it's still something we should address before merging this.

components/user/Tabs/SettingsContent.tsx
dguenther: It sounds like this will be addressed by reloading the page after the update finishes, right?

hooks/useLogin.ts
```diff
@@ -39,78 +32,79 @@ export function useLogin(config: LoginProps = {}) {
...
-
-        // eslint-disable-next-line no-console
-        console.warn('general error!', e)
-      })
+      }
     }
-  }, [$metadata, $setMetadata, redirect, $status, $router])
-
+    loadAndCheck()
+  }, [$metadata, $status, reloadUser])
```
dguenther, line 104: I noticed this `$metadata` in the effect dependencies is causing an extra `/me` fetch to happen. It happens because `$setStatus` gets called before `$setMetadata`, I believe. But since we only want this effect to run once on page load, we could remove the `$metadata` check entirely, and also remove the `$status` check in the `catch` block to get status out of the dependencies as well. That should drop the `/me` calls back down to the same as on production.

dguenther:
I resolved all addressed comments -- there are still some unaddressed issues that are open that have comments on them, and once those are addressed, I'll do another pass on PR review. Thanks!

hooks/useLogin.ts
brekk: @dguenther You mean like:
```
  useEffect(() => {
    const loadAndCheck = async () => {
      try {
        await reloadUser()
      } catch (e) {
        $setStatus(STATUS.FAILED)
      }
    }
    loadAndCheck()
  }, [reloadUser])
```
?

dguenther: Yep, I believe so. That should do ~2 `/me` calls instead of 3.

hooks/usePaginatedEvents.ts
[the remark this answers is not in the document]
brekk: @dguenther I'm a little confused by this comment -- it seems like we were already using all the operators, I just added one level of resolving potentially undefined values here. I understand if your argument is "let's have all impossible states fail", but from my perspective we were already covering a single impossible case, now we're coalescing a definite value regardless of anything. Am I off here?

components/user/Tabs/SettingsContent.tsx
brekk: Ah, I missed the original context. Updated.

hooks/usePaginatedEvents.ts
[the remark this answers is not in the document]
dguenther: When I mentioned "unless TypeScript requires us to," I meant let's only use it if the value has the possibility to be `null` or `undefined`. In this case, `events` has the possibility to be undefined, but if `events` is defined, `events.metadata` will never be undefined (based on the TypeScript types).

Oh, maybe it wasn't clear that the `?.` operator is short-circuiting? The rest of the expression won't be evaluated if a `?.` operator is used on a null/undefined value.

components/user/Tabs/SettingsContent.tsx
```diff
@@ -1,59 +1,260 @@
...
+  const $discord = useField({
+    ...EDITABLE_FIELDS.telegram,
+    defaultValue: _telegram,
+    touched: !!_telegram,
+  })
+  const $telegram = useField({
+    ...EDITABLE_FIELDS.discord,
+    defaultValue: _discord,
+    touched: !!_discord,
+  })
```
dguenther, line 108: I just noticed the Discord and Telegram fields seem to be swapped, so updating one actually updates the other.

brekk: Great catch. Addressed.

hooks/usePaginatedEvents.ts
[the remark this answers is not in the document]
brekk: Ahhhhhh. I did not know that it was a short-circuiting operation. That makes tons of sense.

dguenther (approved):
Looks good!

PR #134, opened 2021-12-10
IRO-1402: Fields can set whitespace controls

brekk:
## Summary

- [x] Fixes IRO-1402 - Trim Graffiti on Registration

Now form fields can specify their `WHITESPACE` rules:

1. `WHITESPACE.DEFAULT` - This keeps the behavior of all extant things
2. `WHITESPACE.TRIMMED` - This implicitly trims (removes trailing + leading whitespace) strings, so if you write `     cooluser     ` (note the trailing spaces) the system only registers `cooluser` as the value.
3. `WHITESPACE.BANNED` - This prevents any whitespace from being entered on a given form.

## Testing Plan

1. Go to: [Sign Up](https://website-testnet-git-trim-yo-strings-ironfish.vercel.app/signup) (you may need to log out first)
2. Try entering spaces on the *Email* or *Social* (Discord / Telegram) fields, they will not be allowed.
3. Spaces can be entered on the *Graffiti* field, they will be allowed to be entered but the resulting value (which will be in the email you receive) won't have leading or trailing spaces, but will be allowed internal to the string boundaries. (e.g. `Blah Blah` is valid, but is the same as `       Blah Blah        `)

## Breaking Change

Is this a breaking change? Not to anything which hasn't changed its `whitespace` value.

### DIFF
--- a/components/Form/TextField.tsx
+++ b/components/Form/TextField.tsx
@@ -64,35 +64,40 @@ export const TextField = ({
   isRadioed,
   disabled,
   autotrim,
-}: Field) => (
-  <LabelledRow
-    key={id}
-    id={id}
-    label={label}
-    valid={valid}
-    errorText={errorText}
-    disabled={disabled}
-  >
-    {isRadioed && options.length > 0 && (
-      <RadioOptions
-        disabled={disabled}
-        groupName={`${id}-group`}
-        options={options}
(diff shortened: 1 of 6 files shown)


deekerno (approved):
Tested, works as intended. 👍🏽

PR #136, opened 2021-12-13
Support local editing with user fixture

brekk:
## Summary

We've made the decision (see previous [Login PR](https://github.com/iron-fish/website-testnet/pull/126#discussion_r764183123)) to make some of our UI hang in a loading state until it hears back from Magic Link. This is good in non-local environments, but locally it's a bit of a PITA to use things.

## Testing Plan

Make sure this doesn't change anything except locally.

## Breaking Change

No, unless you set `NEXT_PUBLIC_LOCAL_USER` to true in production.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,3 +1,4 @@
 NEXT_PUBLIC_API_URL=https://api.ironfish.network
 API_URL=https://api.ironfish.network
 NEXT_PUBLIC_MAGIC_PUBLISHABLE_KEY=test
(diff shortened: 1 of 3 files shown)


leanthebean (approved):

leanthebean (commented):
actually, do we need this PR since this can be solved with derek's suggestion?

brekk:
Gonna merge this for now. We can revisit other options in the future but this will unblock anyone and keep our moving parts low for now.

PR #138, opened 2021-12-14
Linkable FAQ section

leanthebean:
## Summary
Linkable FAQ section

## Testing Plan
tested on local 

## Breaking Change
No

### DIFF
--- a/components/FAQ/QuestionAnswer.tsx
+++ b/components/FAQ/QuestionAnswer.tsx
@@ -1,5 +1,6 @@
 import { Box } from 'components/OffsetBorder/Box'
 import { ReactNode } from 'react'
+import styles from './styles.module.css'
 
 type Props = {
   index: number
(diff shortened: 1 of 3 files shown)


components/FAQ/styles.module.css
```diff
@@ -0,0 +1,8 @@
+.directLink:hover,
+.directLink:focus {
+  opacity: 100%;
+}
+
+.directLink {
+  opacity: 0%;
```
brekk, line 7: `transition: opacity 0.3s ease-out;` might be a classy addition?

components/FAQ/QuestionAnswer.tsx
```diff
@@ -11,7 +12,17 @@ export default function QuestionAnswer({ index, question, answer }: Props) {
...
+          {question}{' '}
+          <a
+            className={styles.directLink}
+            aria-hidden="true"
+            href={'#' + index}
+            title="Direct link to question"
+          >
+            #
+          </a>{' '}
+        </div>
```
brekk, line 25: I might pitch something more URL-friendly / memorable than just numbers here? Maybe something like: [slug](https://www.npmjs.com/package/slug) ?

```js
import slug from 'slug'
// [...] naive implementation
const slugged = slug(question.slice(0, 30))
```
Then replace this bit with
```suggestion
        <div id={slugged} className="font-extended text-[1.75rem] mb-4">
          {question}{' '}
          <a
            className={styles.directLink}
            aria-hidden="true"
            href={'#' + slugged}
            title="Direct link to question"
          >
            #
          </a>{' '}
        </div>
```
B/c in my mind it's much better if we can link something like https://testnet.ironfish.network/faq#what-are-the-minimum-hardware-requirements-for-running-a-node than something numeric, especially if we change the order of things inadvertently.

brekk (changes requested):
Minor suggestion for more URL friendly / human-memorable URLs.

components/FAQ/QuestionAnswer.tsx
brekk: Another, potentially option, upon reflection, would be to add manual ids to things, i.e. `id: "minimum-hardware-requirements"` to the source (that is, the raw `questions` in `page/faq.tsx`)

brekk (approved):
:shipit:

PR #153, opened 2021-12-23
remove old api stuff

brekk:
## Summary

We're being hit by a bot attack, I'm not certain whether it's on the old `api` surface, but since we no longer use that, this _should_ be safe.

[Vercel Docs](https://vercel.com/docs/concepts/functions/serverless-functions#next.js-logs)

I grepped around for usage of `/api` and afaik these are the only things still extant, and I don't think we need them?


rohanjadvani:
Can we remove the `@magic-sdk/admin` package from `package.json`? We shouldn't need that in the UI.

rohanjadvani (approved):
Looks good to me, but I'll wait for @dguenther to take a look as well

dguenther (approved):

PR #154, opened 2021-12-23
Load user page data client-side

dguenther:
## Summary

Updates the user page to fetch data on the client rather than on the server.

## Testing Plan

There isn't much test data on staging anymore, but so far I've tested that the user page loads.

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/hooks/usePaginatedEvents.ts
+++ b/hooks/usePaginatedEvents.ts
@@ -1,22 +1,22 @@
-import { useCallback, useState } from 'react'
+import { Dispatch, SetStateAction, useCallback, useMemo } from 'react'
 import * as API from 'apiClient'
 
 export function usePaginatedEvents(
   userId: string,
   limit: number,
-  initialValue: API.ListEventsResponse
+  events: API.ListEventsResponse | undefined,
+  setEvents: Dispatch<SetStateAction<API.ListEventsResponse | undefined>>
 ): {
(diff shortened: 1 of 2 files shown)


hooks/usePaginatedEvents.ts
```diff
@@ -28,25 +28,24 @@ export function usePaginatedEvents(
...
-    [fetchEvents, $events]
+    async () =>
+      events && fetchEvents(events.data[events.data.length - 1].id, 'after'),
+    [fetchEvents, events]
   )
 
   const fetchPrevious = useCallback(
-    async () => fetchEvents($events[0].id, 'before'),
-    [fetchEvents, $events]
+    async () => events && fetchEvents(events.data[0].id, 'before'),
```
brekk, line 44: Any need for more safety from potentially undefined values here? (not certain from reading)

brekk (commented):
High-level, code looks good; called out maybe a place which needs more safety -- right now I am unable to do anything on Vercel to test.

brekk (approved):
The places I'd expect to still work still do AFAICT, so :shipit:

hooks/usePaginatedEvents.ts
dguenther: It seems like there should potentially be some bounds checking here for `undefined` when pulling values out of the arrays, but since it didn't have it before either, I'm okay shipping this for now.

PR #156, opened 2021-12-23
Remove gitparty from package-scripts

dguenther:
## Summary

gitparty was removed in #155, so this script shouldn't be needed now either.

## Testing Plan

None needed.

## Breaking Change

```
[ ] Yes
[x] No
```


rohanjadvani (approved):

brekk (approved):
Good catch!

PR #159, opened 2022-01-04
Type useLocalLogin

NullSoldier:
## Summary

useLocalLogin returned an inconsitent return type from useLogin. This
fixes that, and moves it into its own file so we have 1 hook per file as
the code base seems to use.

## Testing Plan
Ran code with both hooks

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Tested that "cooldev" shows up when `NEXT_PUBLIC_LOCAL_USER` is set to true.

PR #160, opened 2022-01-04
Cleanup to useLogin hook

NullSoldier:
## Summary
Each commit is a different small cleanup, and I explain each one in the comments of the PR.

## Testing Plan
Will run it locally `NEXT_PUBLIC_LOCAL_USER` as false, and will force various error cases.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -171,7 +171,7 @@ export async function login(email: string): Promise<any> {
 
 export async function getUserDetails(
   token: string
-): Promise<ApiUserMetadata | ApiError | LocalError> {
+): Promise<ApiUserMetadata | LocalError> {
   try {
     const data = await fetch(`${API_URL}/me`, {
       headers: {
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -171,7 +171,7 @@ export async function login(email: string): Promise<any> {
 
 export async function getUserDetails(
   token: string
-): Promise<ApiUserMetadata | LocalError> {
+): Promise<ApiUserMetadata | ApiError | LocalError> {
   try {
     const data = await fetch(`${API_URL}/me`, {
       headers: {
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -180,7 +180,10 @@ export async function getUserDetails(
       },
     })
     return data.json()
-  } catch (e) {
-    return new LocalError(e.message, ENDPOINT_UNAVAILABLE)
+  } catch (e: unknown) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const message = String((e as any)?.message || '')
+
+    return new LocalError(message, ENDPOINT_UNAVAILABLE)
   }
 }
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -180,10 +180,7 @@ export async function getUserDetails(
       },
     })
     return data.json()
-  } catch (e: unknown) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const message = String((e as any)?.message || '')
-
-    return new LocalError(message, ENDPOINT_UNAVAILABLE)
+  } catch (e) {
+    return new LocalError(e.message, ENDPOINT_UNAVAILABLE)
   }
 }
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -9,7 +9,6 @@ import {
   ListLeaderboardResponse,
   MetricsConfigResponse,
   UserMetricsResponse,
-  LoginEvent,
 } from './types'
 import { LocalError } from './errors'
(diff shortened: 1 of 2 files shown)


apiClient/client.ts
```diff
@@ -9,7 +9,6 @@ import {
   ListLeaderboardResponse,
   MetricsConfigResponse,
   UserMetricsResponse,
-  LoginEvent,
```
NullSoldier, line 12: unused

apiClient/client.ts
```diff
@@ -172,7 +171,7 @@ export async function login(email: string): Promise<any> {
 
 export async function getUserDetails(
   token: string
-): Promise<ApiUserMetadata | ApiError | LocalError> {
+): Promise<ApiUserMetadata | LocalError> {
```
NullSoldier, line 174: Unused type in union

apiClient/client.ts
```diff
@@ -181,7 +180,10 @@ export async function getUserDetails(
       },
     })
     return data.json()
-  } catch (e) {
-    return new LocalError(e.message, ENDPOINT_UNAVAILABLE)
+  } catch (e: unknown) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const message = String((e as any)?.message || '')
+
+    return new LocalError(message, ENDPOINT_UNAVAILABLE)
```
NullSoldier, line 187: Cannot access `message` on unknown, and explicitly typed.

hooks/useLogin.ts
```diff
@@ -42,9 +42,9 @@ export function useLogin(config: LoginProps = {}) {
   useEffect(() => {
     const checkLoggedIn = async () => {
       try {
-        // if we're already loaded, quit
         if ($status === STATUS.LOADED) return
-        if (!magic || !magic.user || !magic.user.logout) {
+
+        if (!magic) {
```
NullSoldier, line 47: Unused falsy checks, as the other values cannot be falsy.

hooks/useLogin.ts
```diff
@@ -54,12 +54,13 @@ export function useLogin(config: LoginProps = {}) {
         } catch (error) {}
 
         if (!token) {
-          if (redirect && typeof redirect === 'string') {
+          if (redirect) {
```
NullSoldier, line 57: redirect is a string, so if its truthy it is already proven to TSC that it is a string

hooks/useLogin.ts
```diff
@@ -70,54 +71,44 @@ export function useLogin(config: LoginProps = {}) {
           getUserDetails(token),
         ])
 
-        if ('error' in details || details instanceof LocalError) {
-          $setStatus(STATUS.FAILED)
+        if (details instanceof LocalError) {
```
NullSoldier, line 74: The first check doesn't do anything, instanceof tells TSC that it's the proper type with the right fields.

hooks/useLogin.ts
```diff
@@ -70,54 +71,44 @@ export function useLogin(config: LoginProps = {}) {
...
         }
+
         $setStatus(STATUS.LOADED)
         $setMetadata(details)
         $setMagicMetadata(magicMd)
-      } catch (err) {
-        if (err.toString().indexOf('-32603') > -1) {
+      } catch (err: unknown) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        if ((err as any)?.toString()?.indexOf('-32603') > -1) {
```
NullSoldier, line 92: Access of fields on untyped value. this would crash if `err` was not an object with a `toString()` function.

hooks/useLogin.ts
```diff
@@ -70,54 +71,44 @@ export function useLogin(config: LoginProps = {}) {
...
+
         throw err
       }
     }
+
     if (!$metadata) {
-      try {
-        checkLoggedIn()
-      } catch (e) {
+      checkLoggedIn().catch(e => {
```
NullSoldier, line 102: If you don't await this, then the try/catch doesn't do anything once the first async method starts.

hooks/useLogin.ts
```diff
@@ -70,54 +71,44 @@ export function useLogin(config: LoginProps = {}) {
...
-      if ($status === STATUS.LOADING) {
-        $setStatus(STATUS.FORCED)
-      }
-    }
-    if (timeout > -1) {
-      const tId = setTimeout(forceStatus, timeout)
-      return () => clearTimeout(tId)
-    }
-  }, [$status, $setStatus, timeout])
-  */
```
NullSoldier, line 120: Commented out code

dguenther (approved):
Tested logging in on staging and was able to log in to my staging user.

apiClient/client.ts
NullSoldier: this is actually used

PR #161, opened 2022-01-04
Fix leaderboard watching field causing 2x requests

NullSoldier:
## Summary

This page will watch the entire field from useField()  instead of
the field value, which the form is null, and then not null because
of how it's written in a useEffect(). This cause us to incur 2x
traffic to our API when users load the page. Now we only watch 
the value, which prevents the double request from happening to
the leaderboard.

## Testing Plan

Tested locally

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Verified on the Vercel preview that it only makes one leaderboard request, and tested searching the leaderboard as well.

PR #162, opened 2022-01-04
Remove unused server-side rendering

NullSoldier:
## Summary

Removed the server side rendering that ignored the first search if the results came from the server side props.

## Testing Plan

Tested locally

## Breaking Change

```
[ ] Yes
```


dguenther (approved):
Tested the leaderboard page still loads and searches.

PR #164, opened 2022-01-07
Add snapshot tests

brekk:
## Summary

Add snapshot testing via `jest`.

## Testing Plan

- `nps test` - run tests
- `nps test.snapshot` - update snapshots
- `nps test.watch` - run tests on locally changed files

## Breaking Change

Nein.


__tests__/page-snapshot-user-settings-not-authorized.tsx
```diff
@@ -0,0 +1,35 @@
...
+        query: { id: '111' },
+      },
+    })
+    await waitForElementToBeRemoved(() =>
+      screen.queryByRole('alert', { name: 'Loading' })
+    )
+    expect(container).toMatchSnapshot()
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.warn('cool', e)
```
leanthebean, line 33: if this test fails it'll console log out the message "cool"? if so, can we make it more descriptive?

__tests__/page-snapshot-user-settings.tsx
```diff
@@ -0,0 +1,30 @@
+import User from 'pages/users/[id]'
+import { screen, waitForElementToBeRemoved } from '@testing-library/react'
+import { render } from 'jest.utils'
+import fetch from 'jest-fetch-mock'
+import { mockUserPage } from './page-snapshot-user'
+
+beforeEach(() => {
+  fetch.resetMocks()
```
leanthebean, line 8: why do you need to reset mocks before every test? there's only one test

__tests__/page-snapshot-user-settings.tsx
```diff
@@ -0,0 +1,30 @@
...
+        query: { id: '111' },
+      },
+    })
+    await waitForElementToBeRemoved(() =>
+      screen.queryByRole('alert', { name: 'Loading' })
+    )
+    expect(container).toMatchSnapshot()
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.warn('cool', e)
```
leanthebean, line 28: again, "cool" is not descriptive. we shouldn't be fighting with our linter. looks like you're just doing this to use Error`e` so the linter doesn't yell at you for unused variables? maybe fix that issue instead?

__tests__/page-snapshot-user.tsx
```diff
@@ -0,0 +1,138 @@
...
+    : matches('/events?')
+    ? FIXTURE.events
+    : endsWith('granularity=lifetime')
+    ? FIXTURE.allTimeMetrics
+    : matches('/metrics?granularity=total')
+    ? FIXTURE.weeklyMetrics
+    : matches('/metrics/config')
+    ? FIXTURE.metricsConfig
+    : { error: "You haven't implemented this route yet" }
+  // console.log({ url: req.url, body, init: goodResponse })
```
leanthebean, line 112: remove unused console log comment

components/Loader/index.tsx
```diff
@@ -17,7 +17,12 @@ export const Loader = () => (
         </clipPath>
       </defs>
     </svg>
-    <div className={styles.fishScale}>
+    <div
```
leanthebean, line 20: Looks like this doesn't belong in this PR? 

Let's keep our PRs small and manageable. If we already have a 13k PR that introduces a different paradigm for testing, let's not throw in other changes as well. We can make a separate PR for this.

cypress/integration/pages/about.ts
```diff
@@ -47,21 +47,11 @@ describe('/about', () => {
         text: 'Submit an issue ',
         href: 'https://github.com/iron-fish/ironfish/issues',
       },
-      {
```
leanthebean, line 50: what's happening here? why was this removed? it sounds like this PR's goal is to introduce snapshot testing, and this change makes it look like other changes were made.

__tests__/component-select.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const { container } = selectWithProps({
+    disabled: false,
+    options,
+    defaultValue: 'a',
+    value: 'a',
+    onChange: jest.fn(),
+    setTouched: jest.fn(),
+    className: 'test',
+  })
+  expect(container).toMatchSnapshot()
```
leanthebean, line 21: im confused what this is testing. the values (options) are exactly what is in the snapshot. it's just testing that the snapshot that we ourselves generate has the correct values?

leanthebean (changes requested):
first pass

__tests__/page-snapshot-user-settings.tsx
brekk: Incorrect, the mocks are persisted across runs.

components/Loader/index.tsx
brekk: It belongs in this PR so that we can test against the `role`.

PR #167, opened 2022-01-14
Increase weekly PR points to 5000

dguenther:
## Summary

Updating the "Earn up to 5000 points per week" on the About page was pretty straightforward. For the weekly stats page, since PRs can be awarded variable amounts of points, it doesn't exactly work to show the maximum number of PRs that can be submitted per week -- I changed it to instead show the points as the major value, and the point value max. Then underneath, it shows the instances of each event (i.e. number of blocks mined, number of PRs submitted)

image

## Testing Plan

Load up some users and check that their pages display correctly. You can include more data into the data that's shown in the current week on staging by editing the `getWeeklyStart` function locally.

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/components/user/Tabs/WeeklyContent.tsx
+++ b/components/user/Tabs/WeeklyContent.tsx
@@ -8,18 +8,6 @@ type WeeklyContentProps = {
   metricsConfig: API.MetricsConfigResponse
 }
 
-// Given a metrics config response, computes the maximum number of
-// events for a given metric that can be scored per-week.
-function getWeeklyMetricMax(
-  key: API.EventType,
-  metricsConfig: API.MetricsConfigResponse
-) {
-  return (
-    metricsConfig.weekly_limits[key] /
-    metricsConfig.points_per_category.BLOCK_MINED
-  )
-}
-
 export default function WeeklyContent({
   weeklyMetrics,
   metricsConfig,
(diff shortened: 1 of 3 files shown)


components/user/Tabs/WeeklyContent.tsx
```diff
@@ -29,42 +17,38 @@ export default function WeeklyContent({
       <WeeklyMetricCard
         title="Blocks Mined"
         metric={weeklyMetrics.metrics.blocks_mined}
-        metricValueMax={getWeeklyMetricMax(
-          API.EventType.BLOCK_MINED,
-          metricsConfig
-        )}
+        metricValueMax={metricsConfig.weekly_limits[API.EventType.BLOCK_MINED]}
```
brekk, line 20: Nit, but I'd find it more readable if you did something like:

```
const {
  [EventType.BLOCK_MINED]: _blockMinedMetric
  [EventType.BUG_CAUGHT]: _bugsCaughtMetric
} = metricsConfig.weekly_limits
```

But that's obviously subjective -- I just find the current structure to be not very DRY.

brekk (commented):
Seems like https://website-testnet-git-update-pr-points-ironfish.vercel.app/users/3 should have PRs in the stats? Or am I not testing correctly?

dguenther:
None of the PRs are from the current week, so I wouldn't expect them to show up in the weekly stats. They don't seem to on master either, at least

components/user/Tabs/WeeklyContent.tsx
dguenther: Updated this, let me know if this looks like how you mean!

components/user/Tabs/WeeklyContent.tsx
```diff
@@ -8,63 +8,49 @@ type WeeklyContentProps = {
...
   weeklyMetrics,
   metricsConfig,
 }: WeeklyContentProps) {
+  const {
+    [API.EventType.BLOCK_MINED]: blockMinedLimit,
+    [API.EventType.BUG_CAUGHT]: bugsCaughtWeeklyLimit,
+    [API.EventType.SOCIAL_MEDIA_PROMOTION]: promotionLimit,
+    [API.EventType.PULL_REQUEST_MERGED]: prLimit,
+    [API.EventType.COMMUNITY_CONTRIBUTION]: contributionLimit,
+  } = metricsConfig.weekly_limits
```
brekk, line 21: 🙌 Awesome, much clearer, thanks

brekk:
> None of the PRs are from the current week, so I wouldn't expect them to show up in the weekly stats. They don't seem to on master either, at least

Ah, I see, I wasn't looking at the right tab. Yes, I can see this now.

brekk (approved):

PR #168, opened 2022-01-14
IRO-1420 - week of activity indicator

brekk:
IRO-1420 + IRO-1418

## Summary

CleanShot 2022-01-26 at 15 10 23
CleanShot 2022-01-26 at 15 10 00

## Testing Plan

[Look at it](https://website-testnet-git-feature-iro-1420-week-of-activity-ironfish.vercel.app/users/1#recent-activity), verify the data is accurate and that the new affordances are useful.

## Breaking Change

Nope


pages/users/[id].tsx
```diff
@@ -170,6 +153,7 @@ export default function User({ loginContext }: Props) {
 
   const avatarColor = graffitiToColor($user.graffiti)
   const ordinalRank = numberToOrdinal($user.rank)
+  const signupDate = new Date(2021, 11, 1)
```
Comment (dguenther) [inline pages/users/[id].tsx:156]: Is this intended to be the testnet start date?

components/user/EventRow.tsx
```diff
@@ -0,0 +1,296 @@
...
+      >
+        Week {week}
+      </td>
+    </tr>
+  )
+}
+const weeksBetween = (start: Date, end: Date) =>
+  eachWeekOfInterval({ start, end })
+
+// type Predicate<T> = (x: T) => boolean
```
dguenther, line 235: Safe to delete?

components/user/EventRow.tsx
```diff
@@ -0,0 +1,296 @@
...
+type WeeklyData = {
+  week: number
+  date: Date
+  prior: Date
+  events: ApiEvent[]
+}
+
+export const renderEvents = (start: Date, rawEvents: ApiEvent[]) => {
+  const now = new Date()
+  const weeksThisYear = weeksBetween(start, now)
```
dguenther, line 266: I don't think this is just this year, maybe `weeks`?

components/user/EventRow.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+const makeCounter = () => {
+  let x = 0
+  return () => ++x
+}
+
+const sortEventsByDate = (xs: ApiEvent[]) =>
+  xs.sort((a: ApiEvent, b: ApiEvent): number =>
+    a.occurred_at > b.occurred_at ? 1 : -1
+  )
```
dguenther, line 255: This reverses the order the events are displayed in from how it is currently. I think I'd expect the most recent events to be first, since those are what I'd be most likely to want to check to see if they're logged correctly. But assuming

Aside from that though, if the order is as indended, the pagination logic needs to be changed to handle the new order -- when you click "Next", it loads earlier events rather than later ones.

components/user/EventRow.tsx
```diff
@@ -0,0 +1,296 @@
...
+  } else if (type === EventType.COMMUNITY_CONTRIBUTION) {
+    return <>View contribution</>
+  } else if (type === EventType.SOCIAL_MEDIA_PROMOTION) {
+    return <>Promoted on {hostname}</>
+  }
+  return <>UNHANDLED: {type}</>
+}
+
+const formatEventDate = (d: Date) =>
+  format(d, `do MMM. Y - kk':'mm':'ss`, {
```
dguenther, line 170: This format style is a bit different from what's in the Linear issue, did someone decide to change it?

components/user/EventRow.tsx
```diff
@@ -0,0 +1,296 @@
...
+import ActivitySocial from 'components/icons/ActivitySocial'
+import ChevronRight from 'components/icons/ChevronRight'
+
+import {
+  EventType,
+  ApiEvent,
+  ApiEventMetadata,
+  ApiEventMetadataBlockMined,
+  ApiEventMetadataWithLink,
+} from 'apiClient/index'
```
dguenther, line 28: This can just be `apiClient`

components/user/EventRow.tsx
```diff
@@ -0,0 +1,296 @@
...
+  COPIED = 'copy_copied',
+  RESETTING = 'copy_resetting',
+}
+
+const CopyableHash = ({ hash }: CopyableHashProps) => {
+  const [, $setCopied] = useClipboard(hash, {
+    successDuration: 1000,
+  })
+  const [$hover, $setHover] = useState<CopyState>(CopyState.IDLE)
+  const abbrevHash = hash.slice(45, Infinity)
```
dguenther, line 97: I don't think the `Infinity` is needed here, the second parameter in `slice` is optional

pages/users/[id].tsx
```diff
@@ -251,24 +235,21 @@ export default function User({ loginContext }: Props) {
...
                     </thead>
                     <tbody className="text-sm">
-                      {$events.data.map(e => (
-                        <tr key={e.id} className="border-b border-black">
-                          <td className="py-4">{displayEventType(e.type)}</td>
-                          <td>{new Date(e.occurred_at).toLocaleString()}</td>
-                          <td>{e.points}</td>
-                        </tr>
-                      ))}
+                      {renderEvents(signupDate, $events.data as API.ApiEvent[])}
```
Comment (dguenther) [inline pages/users/[id].tsx:252]: Do you want to propagate the `readonly` or remove it here so it's not necessary to hard-cast it? That way we'll avoid having to update the cast if we change the type of `$events` in the future.

dguenther (changes requested):

components/user/EventRow.tsx
brekk: This makes sense -- I was going primarily from the Figma and didn't consider the delta here. I'll revert / reverse the order.

components/user/EventRow.tsx
brekk: Good call, in an earlier version of the code, the naming was more accurate, but I'll change.

pages/users/[id].tsx
Comment (brekk) [reply pages/users/[id].tsx:156]: It is!

Comment (dguenther) [reply pages/users/[id].tsx:156]: Ah, I think it started on Dec. 1 rather than Nov. 1, but also this should be named something like `startDate` instead

Comment (brekk) [reply pages/users/[id].tsx:156]: I'd be happy to rename the variable -- the `Date` constructs are problematic, but that is in fact December 1, 2021

pages/users/[id].tsx
Comment (brekk) [reply pages/users/[id].tsx]: TIL about this. Added to one function and then a different callsite needed this casting because it does a `.sort`.

components/user/EventRow.tsx
brekk: Reverted

brekk:
@dguenther I _think_ I've addressed all extant feedback -- I noticed that this page isn't currently very usable in smaller breakpoints so I will plan to address that before this goes out.

brekk:
Since it looks like I might need explicit designs for the smaller breakpoints, gonna punt on that effort and have put it in [Linear](https://linear.app/ironfish/issue/IRO-1527/make-user-stats-more-usable-on-small-breakpoints)

components/user/EventRow.tsx
```diff
@@ -0,0 +1,297 @@
...
+  return (
+    weeks
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .reduce((agg: any, date: Date) => {
+        const prev = agg[agg.length - 1]
+        const prior = prev ? prev.date : start
+        const events = sortEventsByDate(
+          eventsBetween(prior, date, rawEvents as ApiEvent[])
+        )
+        const week = counter()
```
dguenther, line 275: I haven't looked at it enough to figure out where the issue is, but it looks like the week count starts on 2 instead of 1, based on this user: https://website-testnet-4mat9grpo-ironfish.vercel.app/users/1

components/user/EventRow.tsx
```diff
@@ -0,0 +1,297 @@
...
+  eachWeekOfInterval({ start, end })
+
+const eventsBetween = (
+  start: Date,
+  end: Date,
+  events: ApiEvent[]
+): ApiEvent[] =>
+  events.filter(e => {
+    const time = parseISO(e.occurred_at)
+    return isAfter(time, start) && isBefore(time, end)
```
dguenther, line 242: I'm not 100% sure that the time bounding is right here -- Looking at this user: https://website-testnet-4mat9grpo-ironfish.vercel.app/users/1  The first event in Week 3 is `2021-12-05 01:00:14` (which I assume is PST), which as UTC is `Sun, 05 Dec 2021 09:00:14 GMT`.

Based on the leaderboard, the weeks are supposed to roll over on Monday at midnight, UTC.

dguenther:
Other than these two comments I just added, looks good!

components/user/EventRow.tsx
brekk: Good catch. I addressed this.

components/user/EventRow.tsx
brekk: I adjusted things after finding a `weekStartsOn` parameter: https://github.com/iron-fish/website-testnet/pull/168/files#diff-e2896aa815d18da702e45b20bd81e18ddf7bf6c79623a7b67a21a0828833da31R237

CleanShot 2022-02-07 at 13 12 37

The only potential outlier is that the week of our first week started on Monday the 29th of December, but I think it's now working correctly.

dguenther: Nice, it's close! The week range still needs to use UTC -- you can check it by looking on https://website-testnet-git-feature-iro-1420-week-of-activity-ironfish.vercel.app/users/1 -- at the end of some weeks, like Week 1 and Week 3, points start getting counted again, and if you take those times and convert them to UTC, they'll be early in the morning on Monday UTC.

brekk:
For reference, I used a combination of the two highest voted answers here: https://stackoverflow.com/questions/58561169/date-fns-how-do-i-format-to-utc to solve the UTC pain points, but open to other options if you have them.

components/user/EventRow.tsx
dguenther: This may have been resolved by mistake, looks like it's still here?

dguenther (approved):
Tested this out on the preview site and looks like it works, nice!

components/user/EventRow.tsx
brekk: Oh good catch, my apologies.

PR #174, opened 2022-02-01
Defining points for testnet categories

leanthebean:
Defining points for promoting the testnet and contributing to the cmmunity categories

## Summary

Screen Shot 2022-02-01 at 10 47 29 AM
(and same for Contributing to the Community section) 

## Testing Plan 
Looked at it on localhost

## Breaking Change

```
[X] No
[ ] Yes
```

### DIFF
--- a/components/About/data.tsx
+++ b/components/About/data.tsx
@@ -109,7 +109,14 @@ export const callsToAction = {
       title: 'Contributing to the Community',
       content:
         'Help us make Iron Fish more accessible to a wider audience! Published tutorials, articles, forum posts, videos, scripts, and translations are examples of contributions that can earn you points.',
-      comingSoon: true,
+      submissionForm: true,
+      points: [
+        'XSmall = 10 points',
+        'Small = 100 points',
+        'Medium = 500 points',
+        'Large = 750 points',
+        'XLarge = 1000 points',
+      ],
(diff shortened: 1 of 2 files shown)


components/About/data.tsx
```diff
@@ -109,7 +109,14 @@ export const callsToAction = {
       title: 'Contributing to the Community',
       content:
         'Help us make Iron Fish more accessible to a wider audience! Published tutorials, articles, forum posts, videos, scripts, and translations are examples of contributions that can earn you points.',
-      comingSoon: true,
+      submissionForm: true,
+      points: [
+        'XSmall = 10 points',
```
brekk, line 114: Minor, not a blocker, but maybe "Extra Small" instead of "XSmall"?

components/About/data.tsx
```diff
@@ -109,7 +109,14 @@ export const callsToAction = {
...
+        'Small = 100 points',
+        'Medium = 500 points',
+        'Large = 750 points',
+        'XLarge = 1000 points',
```
brekk, line 118: Samesies

brekk (approved):
Suggestions but no blockers

dguenther:
I think integration tests need to be updated as well for this.

PR #175, opened 2022-02-01
IRO-1396 - Add weekly countdown timer

brekk:
## Summary

CleanShot 2022-02-07 at 10 49 28@2x

## Testing Plan

See if the countdown timer is useful / provides an affordance

## Breaking Change

Nope

### DIFF
--- a/components/leaderboard/CountdownTimer.tsx
+++ b/components/leaderboard/CountdownTimer.tsx
@@ -1,0 +1,37 @@
+import { useEffect, useState } from 'react'
+import { Box } from 'components/OffsetBorder/Box'
+import { intervalToDuration, nextMonday, set } from 'date-fns'
+import type { Duration } from 'date-fns'
+
+const customFormatDuration = (x: Duration) =>
+  `${x.days ? x.days + 'd : ' : ''}${x.hours ? x.hours + 'hr : ' : ''}${
+    x.minutes ? x.minutes + 'min : ' : ''
+  }${x.seconds ? x.seconds + 'sec' : '0sec'}`
+
+const CountdownTimer = () => {
+  const now = new Date()
+  const [$time, $setTime] = useState(now)
+  const end = nextMonday(
+    set(now, { hours: 0, minutes: 0, seconds: 0, milliseconds: 0 })
+  )
+  useEffect(() => {
+    const i = setInterval(() => $setTime(new Date()), 3e3)
+    return () => clearInterval(i)
+  }, [])
+  const ii = intervalToDuration({ start: $time, end })
+  return (
+    <div className="w-full flex flex-col justify-center items-center text-center">
+      <Box className="w-[18.25rem]" background="bg-ifpink">
+        <div className="w-[18.25rem] px-4 py-2 m-auto">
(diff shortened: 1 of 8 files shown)


components/leaderboard/CountdownTimer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }${x.seconds ? x.seconds + 'sec' : '0sec'}`
+
+const CountdownTimer = () => {
+  const now = new Date()
+  const [$time, $setTime] = useState(now)
+  const end = nextMonday(
+    set(now, { hours: 0, minutes: 0, seconds: 0, milliseconds: 0 })
+  )
+  useEffect(() => {
+    const i = setInterval(() => $setTime(new Date()), 3e3)
```
NullSoldier, line 18: This looks really laggy with 3000 ms, it should be a lot lower. Ideally it should use requestAnimationFrame but setInterval should be fine since we only show 1 second intervals and we don't need less than that.

Also why is this using scientific notation? Can we just use normal numbers that people are used to?

```suggestion
    const i = setInterval(() => $setTime(new Date()), 1000)
```

brekk: I find scientific notation very scannable. I've adjusted to 1ms cycles

pages/leaderboard.tsx
```diff
@@ -143,8 +146,9 @@ export default function Leaderboard({ loginContext }: Props) {
               'md:py-5',
               'md:px-4'
             )}
-            buttonLink={!isLoggedIn ? '/signup' : ''}
-          />
+          >
+            {isLoggedIn && <CountdownTimer />}
```
dguenther, line 150: Is it important that this only shows up on when logged in? It seems like it could be relevant to non-logged-in users as well -- they could still be testnet participants that haven't logged in yet.

components/leaderboard/CountdownTimer.tsx
```diff
@@ -0,0 +1,43 @@
...
+  `${x.days ? x.days + 'd : ' : ''}${x.hours ? x.hours + 'hr : ' : ''}${
+    x.minutes ? x.minutes + 'min : ' : ''
+  }${x.seconds ? x.seconds + 'sec' : '0sec'}`
+
+const CountdownTimer = () => {
+  const now = new Date()
+  const [$time, $setTime] = useState(now)
+  const end = nextMonday(
+    set(now, { hours: 0, minutes: 0, seconds: 0, milliseconds: 0 })
+  )
```
dguenther, line 16: It looks like the countdown is displaying the time until midnight PST instead of midnight UTC -- there probably needs to be some UTC conversion done here or somewhere.

dguenther (changes requested):

pages/leaderboard.tsx
brekk: This was a discussed consideration but the designs were around showing the sign up button for not-logged-in users and the countdown timer for logged in users. I don't feel super strongly (and we had both early on), but this was done with intention.

dguenther (approved):
Looks good!

PR #179, opened 2022-02-08
IRO-1533 -  Bug: Some user events swallowed

brekk:
## Summary

Addresses the issues raised in [Slack](https://iron-fish.slack.com/archives/C02PV6KFPN2/p1644352080469069?thread_ts=1644351745.028479&cid=C02PV6KFPN2).

- [x] - Problem 1: The event grouping made the most recent group only be from Monday at Midnight UTC to "right now" (not UTC). Also guessing that very recently created values were then outside the 8 hour range from PST to UTC. Now the end boundary is the same code which comes from the recent countdown timer (so I refactored both to use the same dependency).
- [x] - Problem 2: Weekly Points weren't rendering correctly: See [PR 180](https://github.com/iron-fish/website-testnet/pull/180) for more details.

## Testing Plan

Verify a user on staging can be credited and that the page reflects that.

## Breaking Change

Nope, it's a fixing change.


brekk:
Did some testing, doesn't seem fixed yet. Gonna keep testing.

apiClient/types.ts
```diff
@@ -22,8 +22,8 @@ export type ApiEventMetadata =
 
 export type ApiEvent = {
   id: number
-  created_at: string
-  updated_at: string
+  created_at?: string
+  updated_at?: string
```
NullSoldier, line 26: These can be deleted as the API never sends these down. https://github.com/iron-fish/ironfish-api/blob/master/src/events/utils/event-translator.ts#L10-L18

utils/date.ts
```diff
@@ -17,19 +24,27 @@ export const weeksBetween = (start: Date, end: Date) => {
...
+): ApiEvent[] => events.filter(withinBounds(start, end))
 
 export const formatInTimeZone =
   (timeZone: string) => (date: Date, formatString: string) =>
     format(utcToZonedTime(date, timeZone), formatString, { timeZone })
 export const formatUTC = formatInTimeZone('UTC')
 export const formatEventDate = (d: Date) =>
   formatUTC(d, `y'-'MM'-'dd HH':'mm':'ss`)
+
+export const nextMondayFrom = (now: Date) => {
```
NullSoldier, line 45: `now` can be any time now right, since this code was factored out?

brekk: Yeah, it can -- would you prefer a different parameter name?

apiClient/types.ts
brekk: Removed

utils/date.ts
NullSoldier: sort of, it just seemed like the code moved and the name just wasn't changed since that variable used to be `now` but isn't anymore. I'd say something like `time` or `from` but I don't think it matters that much. this is just a nit more than anything

leanthebean:
I pulled your branch, changed .env.local file to point to production API instead, and messed with chrome CORS settings to get production api data in localhost, and the bug still persists: 
Screen Shot 2022-02-08 at 10 32 43 PM

brekk:
> I pulled your branch, changed .env.local file to point to production API instead, and messed with chrome CORS settings to get production api data in localhost, and the bug still persists (in the header summary): Screen Shot 2022-02-08 at 10 32 43 PM

@leanthebean When you say "header summary" you mean the Weekly Points being 0? I am happy to investigate that but I believe it is a distinct issue from this bug.

leanthebean:
I clicked around and it looks like the issues i was seeing are gone, but i didn't look at the date-time calculations too closely

pages/users/[id].tsx
```diff
@@ -154,11 +159,12 @@ export default function User({ loginContext }: Props) {
   const avatarColor = graffitiToColor($user.graffiti)
   const ordinalRank = numberToOrdinal($user.rank)
   const startDate = new Date(2021, 11, 1)
+  const endDate = nextMondayFrom(nextMonday(new Date()))
```
Comment (NullSoldier) [inline pages/users/[id].tsx:162]: It took me awhile to understand why this was happening. It gets the next Monday in the local time zone, and then gets the next Monday from that in the UTC time zone, I imagine this it to work around the fact that something can be past UTC-0 weekly cut off so it adds an extra week for safety?

NullSoldier (approved):

NullSoldier:
It just looks like the tests are failing now.

NullSoldier:
I accidentally closed the PR when I was trying to click comment.

It looks like the tests are failing but beyond that it's good to go.

pages/users/[id].tsx
Comment (brekk) [reply pages/users/[id].tsx:162]: Exactly; it's a little non-intuitive but it seems to work correctly

brekk:
> I accidentally closed the PR when I was trying to click comment.
> 
> It looks like the tests are failing but beyond that it's good to go.

Yeah unfortunately I haven't had time to figure that out yet, but all tests fail currently.

Will plan on merging this and then addressing tests shortly.

PR #182, opened 2022-02-09
Fix text overflow UI issue

wd021:
Fixing issue iron-fish/website-testnet#185

Reducing font-size of class '5xl' so `1,000 contributions` doesn't overflow the container.

### DIFF
--- a/components/user/MetricCard.tsx
+++ b/components/user/MetricCard.tsx
@@ -23,9 +23,9 @@ export default function MetricCard({
         <div className=" mb-4 whitespace-nowrap">{title}</div>
         <div className="flex gap-4">
           <div className="text-5xl">{value}</div>
-          <div className="font-favorit mt-1">
+          <div className="font-favorit mt-1 overflow-hidden">
             <div className="leading-tight">{subValueTop}</div>
-            <div className="text-ifsubtextgray leading-tight whitespace-nowrap	">
+            <div className="text-ifsubtextgray leading-tight truncate">
               {subValueBottom}
             </div>
           </div>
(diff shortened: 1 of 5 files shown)


brekk:
@wd021 Hey, thanks for contributing.

I don't think this is quite the right solution, as it has effects beyond just the scope of the mentioned issue.

wd021:
@brekk 

hmm, i was seeing that the 5xl class was only being used there. thought it would be a sufficient solution to reduce the font-size rem by a bit so it doesn't overflow at max 1,000.

how about if i change the solution to truncating the text like so:
Screenshot 2022-02-09 at 21 33 27

brekk:
@wd021 I think the solution could be two-fold -- This is the design Skylar came up with (though the copy is off currently):

CleanShot 2022-02-09 at 14 27 39

But I was thinking `s/contributions/additions/g` to start (or maybe "changes" instead ?), and then the ellipsis was the other direction I was thinking about.

wd021:
gotcha. Well with this new design, I guess the overflow of 1,000 won't be an issue. 

'Changes' might be more accurate than 'additions', but I guess that's just nitpicking.

wd021:
👍 I'm looking to contribute more to Ironfish!

For now I've made these updates
- revert fontSize change
- `s/contributions/changes/g`
- add text ellipsis on overflow

dguenther:
We switched away from the design shown above ("points" rather than individual items like "contributions") because the weekly limits are based on the number of points gained rather than the number of events created. In other words, since PRs can be worth a variable amount of points, there isn't a limit of 1 PR per week, but there's a limit of the number of points gained by submitting PRs.

I actually think it should be "Contributions" as it was originally, since this category is for things like guides written by the community or popular social media posts made by community members -- there isn't necessarily something being changed. I'll roll this back myself to save you some time, then this looks good, thanks!

PR #183, opened 2022-02-11
IRO-1541 - Render events whose occurred_at value is before Testnet started

brekk:
## Summary

Events from before the testnet are valid and should be rendered.

Now they are.

## Testing Plan

1. Add events to a user before Dec. 1, 2021
2. They will render like this: CleanShot 2022-02-11 at 09 36 11
3. Profit ?

## Breaking Change

Nope.

### DIFF
--- a/utils/events.ts
+++ b/utils/events.ts
@@ -20,31 +20,25 @@ export const eventsBetween = (
   events: ApiEvent[]
 ): ApiEvent[] => events.filter(withinBounds(start, end))
 
-export function uniqueById(x: readonly ApiEvent[]) {
-  return x.reduce((agg: ApiEvent[], i: ApiEvent) => {
-    if (!agg.map(({ id }) => id).includes(i.id)) {
-      return agg.concat(i)
-    }
-    return agg
-  }, [])
+export const getAllVisibleEvents = (weeklyData: WeeklyData[]) => {
+  const m = new Map()
+  const allEvents = weeklyData.flatMap(({ events }) => events)
+  allEvents.forEach(e => m.set(e.id, e))
+  return m
 }
-
-export const getAllVisibleEvents = (weeklyData: WeeklyData[]) =>
-  weeklyData.reduce((agg: ApiEvent[], { events }) => {
-    return agg.concat(events)
-  }, [])
 
 export const findMissingEvents = (
   allEvents: readonly ApiEvent[],
   weeklyData: WeeklyData[]
-) =>
-  allEvents.reduce((agg: ApiEvent[], x: ApiEvent) => {
-    const onlyMatched = getAllVisibleEvents(weeklyData)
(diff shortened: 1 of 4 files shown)


utils/events.ts
```diff
@@ -0,0 +1,103 @@
...
+  return isAfter(time, start) && isBefore(time, end)
+}
+export const eventsBetween = (
+  start: Date,
+  end: Date,
+  events: ApiEvent[]
+): ApiEvent[] => events.filter(withinBounds(start, end))
+
+export function uniqueById(x: readonly ApiEvent[]) {
+  return x.reduce((agg: ApiEvent[], i: ApiEvent) => {
```
dguenther, line 24: Do you want to just build a map with `id` as the key here, then return `.values()`? That seems like it would avoid the need to build an intermediate array and loop over it to check uniqueness.

utils/events.ts
```diff
@@ -0,0 +1,103 @@
...
+      return agg.concat(i)
+    }
+    return agg
+  }, [])
+}
+
+export const getAllVisibleEvents = (weeklyData: WeeklyData[]) =>
+  weeklyData.reduce((agg: ApiEvent[], { events }) => {
+    return agg.concat(events)
+  }, [])
```
dguenther, line 35: `flatMap` here might be a bit cleaner.

utils/events.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export const findMissingEvents = (
+  allEvents: readonly ApiEvent[],
+  weeklyData: WeeklyData[]
+) =>
+  allEvents.reduce((agg: ApiEvent[], x: ApiEvent) => {
+    const onlyMatched = getAllVisibleEvents(weeklyData)
+    if (onlyMatched.map(({ id }: { id: number }) => id).includes(x.id)) {
```
dguenther, line 43: The construction of the `onlyMatched` map can just be done once before the `reduce`. You could just have `getAllVisibleEvents` return a map or a set to make the existence check easier too.

dguenther (changes requested):
Looks like a good fix overall, I checked out `coolcoolcool` on staging. Just a few data structure-related change requests then this should be good to go 👍

utils/events.ts
brekk: Realized this function isn't actually in use. So I removed it.

utils/events.ts
brekk: Good call, `flatMap` falls into the category of addendums-to-JS-which-I-recognize-but-rarely-recall

utils/events.ts
brekk: Cleaned it up a bunch and refactored to use `Map` internally.

brekk:
Fixes IRO-1541

(Linear integration)

dguenther (approved):

Issue #184, opened 2022-02-07
Testnet Leaderboard mobile sizing is broken


sternhenri:
I am seeing the following on mobile (iOS in Safari, Firefox, Chrome). Namely, some of the website resizing for mobile compresses the points together with the usernames making it especially hard to understand when usernames end in digits. 

![IMG_3019](https://user-images.githubusercontent.com/3359083/152854717-41988b1d-cf15-4e80-987a-e625057df4a5.PNG)

AmberKiso:
Validated the leaderboard displays correctly on an Android phone, Chrome and Firefox.
If someone validates on Safari, I think we can close this.

dguenther:
I can validate this works on Safari mobile. I'll close this for now, and if anyone runs into mobile issues on the updated version, feel free to open new issues to address those. Glad to get this addressed!

PR #187, opened 2022-02-16
fix leaderboard page on mobile

wd021:
## Summary
Fixing up issue referenced at https://github.com/iron-fish/website-testnet/issues/184. Changes made in this PR
- add extra padding in `<Select>` so text doesn't overlap with arrow on mobile
- `<Select>` arrow height was a bit off
- Leaderboard looks much better on mobile now

before:
Screen Shot 2022-02-16 at 4 02 01 PM

after:
Screen Shot 2022-02-16 at 4 02 06 PM

## Breaking Change

```
[ ] Yes
[x] No
```

### DIFF
--- a/components/Form/Select.module.css
+++ b/components/Form/Select.module.css
@@ -1,6 +1,7 @@
 .customSelectWrapper {
   width: 100%;
   position: relative;
+  padding-right: 1rem;
 }
 .customSelect {
   appearance: none;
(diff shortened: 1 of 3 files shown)


brekk:
@wd021 Hey, thanks for submitting this!

In general I think this is looking pretty good -- I was wondering if you could adjust things so that this gap:
154306820-ca5a3251-4c62-45c3-9ecd-3e123200a469
is shortened per row?

wd021:
@brekk yup, done!

PR #190, opened 2022-02-23
IRO-1528 - Responsify /users/id

brekk:
## Summary

Responsify the User Details page

See [Figma](https://www.figma.com/file/E04zlsQJHbywfHq7zGqwt1/Testnet-Rewards-Page?node-id=8948%3A4508) (though bear in mind that we decided MetricCards in the 832 - 1136 range should have 1/3 width).

## Testing Plan

👀

## Breaking Change

```
[ ] Yes
[x] Nope
```


pages/users/[id].tsx
```diff
@@ -167,8 +187,12 @@ export default function User({ loginContext }: Props) {
   ).toLocaleString()
   const weeklyPoints = $weeklyMetrics.points.toLocaleString()
 
+  const tweetText = `Iron Fish Incentivized Testnet: ${
```
Comment (brekk) [inline pages/users/[id].tsx:190]: Feedback definitely wanted on this copy.

components/Navbar/index.tsx
```diff
@@ -74,17 +74,14 @@ function Navbar({
...
-          // lg - 1024px
-          'lg:text-[18px]',
           // 2lg - 1152px
-          '2lg:px-10',
-          '2lg:text-xl',
+          // '2lg:px-10',
+          // '2lg:text-md',
           // 1440
-          '1.5lg:px-16',
+          // '1.5xl:px-16',
```
brekk, line 84: Awaiting some feedback from @skylarrichard123 on this, but will clean up after.

components/user/EventRow.module.css
```diff
@@ -47,3 +47,15 @@
   opacity: 0 !important;
   display: none !important;
 }
+
+.truncatedHash {
+  direction: rtl;
```
brekk, line 52: This is a [tricky trick](https://davidwalsh.name/css-ellipsis-left)

deekerno (approved):

components/user/EventRow.module.css
dguenther: oh that's pretty cool, nice!

components/user/FishAvatar.tsx
```diff
@@ -1,52 +1,30 @@
 import React from 'react'
+import clsx from 'clsx'
+// import styles from './FishAvatar.module.css'
```
dguenther, line 3: Extra comment here

pages/users/[id].tsx
```diff
@@ -61,6 +76,11 @@ export default function User({ loginContext }: Props) {
     API.MetricsConfigResponse | undefined
   >(undefined)
   const [$fetched, $setFetched] = useState(false)
+  useEffect(() => {
+    if (typeof window !== 'undefined') {
+      $setDomain(`https://` + window.location.hostname)
+    }
+  }, [])
```
Comment (dguenther) [inline pages/users/[id].tsx:83]: Maybe this is okay to inline outside of the `useEffect` hook? It could also just be set as a constant as well.

pages/users/[id].tsx
Comment (dguenther) [reply pages/users/[id].tsx:190]: I'm not sure about the hashtags beyond `#ironfish`, but otherwise seems pretty simple/unopinionated so people can customize it at will, which I like

pages/users/[id].tsx
```diff
@@ -177,51 +201,131 @@ export default function User({ loginContext }: Props) {
...
+                      'font-extended',
+                      'text-3xl',
+                      'md:text-6xl',
+                      'mt-2',
+                      'md:mt-6',
+                      'mb-4',
+                      'max-w-[14rem]',
+                      'md:max-w-[48rem]',
+                      'max-h-[4rem]',
+                      'overflow-hidden'
```
Comment (dguenther) [inline pages/users/[id].tsx:249]: I think this can clip graffitis in the tablet view -- I'm not sure what the best UX is here, but maybe just an ellipsis on overflow would clear up that it's truncated?
image

dguenther (approved):
Looks good! I think the graffiti overflow might be worth fixing up, but otherwise comments here are all minor, so feel free to defer them if you'd like 👍

pages/users/[id].tsx
Comment (brekk) [reply pages/users/[id].tsx]: Yeah, fair call. I elected to use a static string now. I wanted initially to verify how it works.

pages/users/[id].tsx
Comment (brekk) [reply pages/users/[id].tsx]: For this one I increased the possible width and also added an ellipsis.

PR #194, opened 2022-03-01
IRO-1529 Search and Filter

brekk:
## Summary

Adjusted the Search + Filter functionality on the leaderboard to be more helpful at small sizes.

CleanShot 2022-03-02 at 09 33 08

## Testing Plan

👀 

## Breaking Change

```
[ ] Yes
[x] Nope
```

### DIFF
--- a/pages/leaderboard.tsx
+++ b/pages/leaderboard.tsx
@@ -218,6 +218,7 @@ export default function Leaderboard({ loginContext }: Props) {
                 'flex',
                 'items-center',
                 'mb-4',
+                'md:mb-8',
                 'md:w-1/2'
               )}
             >
--- a/pages/leaderboard.tsx
+++ b/pages/leaderboard.tsx
@@ -293,7 +293,7 @@ export default function Leaderboard({ loginContext }: Props) {
                 <div className={clsx('flex-1', 'hidden', 'md:inline')}>
                   USERNAME
                 </div>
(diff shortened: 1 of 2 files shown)


pages/leaderboard.tsx
```diff
@@ -149,44 +157,145 @@ export default function Leaderboard({ loginContext }: Props) {
...
-                <div className="w-16 sm:w-24 hidden md:inline">RANK</div>
-                <div className="flex-1 hidden md:inline">USERNAME</div>
-                <div className="ml-2  hidden md:inline">TOTAL POINTS</div>
+                <div className={clsx('w-16', 'sm:w-24', 'hidden', 'md:inline')}>
+                  RANK
+                </div>
+                <div className={clsx('flex-1', 'hidden', 'md:inline')}>
+                  USERNAME
+                </div>
+                <div className={clsx('ml-2', '', 'hidden', 'md:inline')}>
```
dguenther, line 296: The extra space can probably be removed here, unless it's doing something I don't realize.

dguenther (approved):
Looks good! 🚢

pages/leaderboard.tsx
brekk: Good catch, I wrote a vim script to convert raw `className` strings to `clsx` entries, but it doesn't currently catch double-spaces. Will amend.

PR #202, opened 2022-03-08
IRO-1529 - search / filter redux

brekk:
## Summary

This change didn't make it in the last release (because the underlying HTML tree changed in a way that made these changes not be applied correctly)

CleanShot 2022-03-07 at 16 20 46

## Testing Plan

👀 

## Breaking Change

```
[ ] Yes
[x] Nope
```


dguenther (approved):
Looks good to me!

PR #203, opened 2022-03-08
Fix login button & add `/me`

brekk:
## Summary

If you are on someone else's user page, and you click the button to go to your page, it causes a weird re-render (and re-requests data).

This PR aims to fix that bug by making the button go to a new route: `/me` which is the same as `/users/[your-id]`.

## Testing Plan

👀 

## Breaking Change

```
[ ] Yes
[x] No?
```


brekk:
I think this idea is interesting but likely more work than it's worth. Closing for now

PR #205, opened 2022-03-08
IRO-1399 - Leaderboard Pagination

brekk:
## Summary

Make `/leaderboard` data paginated.

## Testing Plan

👀 

## Breaking Change

```
[ ] Yes
[x] Nope
```

### DIFF
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -6,7 +6,7 @@ import {
   ApiError,
   ApiUser,
   ListEventsResponse,
-  ListLeaderboardResponse,
+  PaginatedListLeaderboardResponse,
   MetricsConfigResponse,
   UserMetricsResponse,
 } from './types'
@@ -72,15 +72,21 @@ export async function updateUser(id: number, partial: PartialUser) {
   return await res.json()
 }
 
-export async function listLeaderboard({
+export async function listUsers({
   search,
   country_code: countryCode,
   event_type: eventType,
+  limit,
+  after,
+  before,
 }: {
   search?: string
   country_code?: string
   event_type?: string
-} = {}): Promise<ListLeaderboardResponse | ApiError> {
+  limit?: number
+  after?: number
+  before?: number
+}): Promise<PaginatedListLeaderboardResponse | ApiError> {
   const params = new URLSearchParams({
     order_by: 'rank',
   })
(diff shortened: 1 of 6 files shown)


pages/leaderboard.tsx
```diff
@@ -64,6 +65,8 @@ const FIELDS = {
 }
 const CTA = `Our incentivized testnet leaderboard shows you who the top point getters are for all-time score, miners, bug catchers, net promoters, node hosting, and more! Click someone’s user name to see a breakdown of their activity. Points are earned during weekly cycles which begin Monday, 12:00am UTC and end Sunday 11:59pm UTC.`
 
+const PAGINATION_LIMIT = 3
```
brekk, line 68: This is set temporarily, it needs to be changed to 25 _before_ this merges.

brekk:
Sounds like there's a bug on the API side regarding `before` sorting, keeping this in Draft but it's ready for review.

apiClient/client.ts
```diff
@@ -72,15 +72,21 @@ export async function updateUser(id: number, partial: PartialUser) {
...
+  before,
 }: {
   search?: string
   country_code?: string
   event_type?: string
-} = {}): Promise<ListLeaderboardResponse | ApiError> {
+  limit?: number
+  after?: number
+  before?: number
+}): Promise<ListLeaderboardResponse | ApiError> {
```
dguenther, line 89: It'd be nice to rename the response type to `ListUsersResponse` if we're going to rename the function (I agree renaming the function makes sense)

pages/users/[id].tsx
```diff
@@ -9,7 +9,7 @@ import Footer from 'components/Footer'
 import Navbar from 'components/Navbar'
 import Loader from 'components/Loader'
 import { Container as OffsetBorderContainer } from 'components/OffsetBorder'
-import EventsPaginationButton from 'components/user/EventsPaginationButton'
+import EventsPaginationButton from 'components/PaginationButton'
```
Comment (dguenther) [inline pages/users/[id].tsx:12]: This could be renamed to `PaginationButton` like on the leaderboard page, to match up the names.

dguenther (approved):
Sounds like we'll wait to ship to master until the `before` issue with `order_by=rank` is fixed on the API, but otherwise looks good!

PR #215, opened 2022-03-14
Fix fish avatar color

brekk:
## Summary

1. https://testnet.ironfish.network/users/104 was throwing errors in the `graffitiToColor` function.
2. I resurrected the `jest` tests and wrote a test for the function.
3. I fixed the function and updated the tests.

## Testing Plan

1. `nps care`
2. Updated the github workflow, so it should run automatically now.

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and what additional work is required, if any.

```
[ ] Yes
[x] No
```


tsconfig.json
```diff
@@ -13,9 +13,17 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
+    "incremental": true,
     "jsx": "preserve",
-    "types": ["cypress"]
+    "types": ["@types/testing-library__jest-dom"]
```
dguenther, line 18: Are the cypress types still needed? I'm not too familiar with how the typeloading works for multiple testing libraries.

dguenther (changes requested):
I think the problem here is that the API is 404ing for that user, and we're not handling that case in the frontend -- The types for the graffitiToColor function indicate it should take a string, but it's not receiving a string.

We should be handling the error on the frontend and displaying an error page, otherwise I expect this will probably bubble up and cause issues in other functions as well, or if not, it will display a page with empty data and no explanation.

I don't mind bringing back the tests if you'd like to do that, but I'd rather not add a default argument to `graffitiToColor`, since it could hide a place where we're actually intending to pass a graffiti value to the function.

tsconfig.json
brekk: Unfortunately yes, they are -- There's another tsconfig which now lives in `/cypress` as well.

brekk:
@dguenther Addressed your feedback, I changed it so that the length defaults to zero rather than defaulting the inputs

tsconfig.json
dguenther: Oh that's fine, I just missed that TSConfig

dguenther:
I think maybe I didn't explain myself well -- this PR doesn't fix this issue mentioned in the description:

> 1. https://testnet.ironfish.network/users/104 was throwing errors 

If you go to https://website-testnet-git-fix-fish-avatar-color-ironfish.vercel.app/users/200, it throws a different error in a different place. The bug that needs to be fixed here is the types for the API responses aren't correct, so we're not handling the `404` from the backend. It's not necessary to make changes to the `graffitiToColor` function -- the types for that function are correct as-is.

brekk:
Killing this PR as I didn't understand the root causes here -- https://github.com/iron-fish/website-testnet/pull/216 is the new one.

PR #216, opened 2022-03-15
handle unknown users

brekk:
## Summary

If you go to a user ID which doesn't exist, our frontend didn't deal with the missing data correctly:

e.g. testnet.ironfish.network/users/100000

## Testing Plan

Go to `/users/[some-number-which-doesn't-exist-as-a-user-id]`

## Breaking Change

No, it's a fixing change.

### DIFF
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -5,6 +5,7 @@ import {
   ApiUserMetadata,
   ApiError,
   ApiUser,
+  GenericApiError,
   ListEventsResponse,
   ListLeaderboardResponse,
   MetricsConfigResponse,
@@ -206,4 +207,9 @@ export async function getUserDetails(
   } catch (e) {
     return new LocalError(e.message, ENDPOINT_UNAVAILABLE)
   }
+}
+
+export const isGenericError = (x: GenericApiError) => {
+  const { statusCode } = x
+  return statusCode && statusCode > 400
 }
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -209,7 +209,9 @@ export async function getUserDetails(
   }
 }
 
-export const isGenericError = (x: GenericApiError) => {
+export const isGenericError = (
+  x: Record<string, unknown>
+): x is GenericApiError => {
   const { statusCode } = x
-  return statusCode && statusCode > 400
+  return typeof statusCode === 'number' && statusCode > 400
 }
--- a/pages/users/[id].tsx
+++ b/pages/users/[id].tsx
@@ -108,10 +108,15 @@ export default function User({ loginContext }: Props) {
 
         if (
           'error' in user ||
+          API.isGenericError(user as API.GenericApiError) ||
           'error' in events ||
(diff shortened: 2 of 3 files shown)


apiClient/client.ts
```diff
@@ -207,3 +208,8 @@ export async function getUserDetails(
     return new LocalError(e.message, ENDPOINT_UNAVAILABLE)
   }
 }
+
+export const isGenericError = (x: GenericApiError) => {
```
dguenther, line 212: You can define the function like this:

```ts
export const isGenericError = (
  x: Record<string, unknown>
): x is GenericApiError => {
  const { statusCode } = x
  return typeof statusCode === 'number' && statusCode > 400
}
```

Then it'll avoid the need to do `as API.GenericApiError` when passing messages into `API.isGenericError`. (As an FYI, `Record<string, unknown>` is generally equivalent to `object`)

pages/users/[id].tsx
```diff
@@ -108,10 +108,15 @@ export default function User({ loginContext }: Props) {
...
+          API.isGenericError(user as API.GenericApiError) ||
           'error' in events ||
+          API.isGenericError(events as API.GenericApiError) ||
           'error' in allTimeMetrics ||
+          API.isGenericError(allTimeMetrics as API.GenericApiError) ||
           'error' in weeklyMetrics ||
-          'error' in metricsConfig
+          API.isGenericError(weeklyMetrics as API.GenericApiError) ||
+          'error' in metricsConfig ||
+          API.isGenericError(metricsConfig as API.GenericApiError)
```
Comment (brekk) [inline pages/users/[id].tsx:119]: This construct isn't my favorite I've ever written -- if there's a common and easier approach I'm happy to try it -- I spend a while messing around with `ApiError` but found it kept complaining.

pages/users/[id].tsx
```diff
@@ -154,7 +159,12 @@ export default function User({ loginContext }: Props) {
 
   // Recent Activity hooks
   const { $hasPrevious, $hasNext, fetchPrevious, fetchNext } =
-    usePaginatedEvents(userId, EVENTS_LIMIT, $events, $setEvents)
+    usePaginatedEvents(
+      userId,
+      EVENTS_LIMIT,
+      $events as API.ListEventsResponse,
```
Comment (dguenther) [inline pages/users/[id].tsx:165]: I think this `as` is removable, the build seems to pass without it.

In general, we should avoid using `as` where possible since it punches a hole in the type system. (you could do `as number` even though it's not actually that)

dguenther (changes requested):

apiClient/client.ts
brekk: Nice, the `x is GenericApiError` is new to me -- I tried the Record in an earlier permutation.

pages/users/[id].tsx
Comment (brekk) [reply pages/users/[id].tsx]: It was required in the very first version, but now it is passing without this cast.

brekk:
@dguenther Thanks for the feedback, your construct is much nicer / more sane. I've updated the branch

dguenther (approved):
Looks good!

PR #219, opened 2022-04-04
Phase 2: About Page

brekk:
## Summary

1. Updated the CallToAction component
   - added a `status` field
   - added a `disabled` field
2. /about page
   - page color is beige now!
   - Added Phase component
   - general vertical spacing
3. Global changes
   - Updated menu to show a notification above the menu
   - Updated menu flyouts to move down correctly when notification is visible
4. CustomBox component changes
   - Updated to allow for the highest possible customization

## Testing Plan

👀 

## Breaking Change

```
[ ] Yes
[x] No
```


pages/about.tsx
```diff
@@ -79,22 +82,29 @@ export default function About({ loginContext }: AboutProps) {
...
-          title="About the Incentivized Testnet"
-          text="Sign up for the Iron Fish incentivized testnet to help make Iron Fish great 💖. Participate to earn testnet points (see Testnet Guidelines below for more details)."
+          title={
+            <>
+              Incentivized Testnet
+              <br />
+              Welcome to Phase 2.
+            </>
+          }
+          text="Get ready to claim your share of 315,000 $IRON! To make Phase 2 fair, we’ve removed weekly limits on point earnings. We’re also introducing a new leaderboard; but don’t worry, all accumulated Phase 1 points will still count to your overall score. You can earn $IRON in a few different ways:"
```
dguenther, line 107: I think somewhere else this "315,000" number was computed from the phase + pool data, should we do that here too?

dguenther (changes requested):
One more thing that I couldn't attach directly to a line of code:

* Does the cypress failed screenshot need to be removed?

brekk:
> * Does the cypress failed screenshot need to be removed?

I didn't think it needed to be part of the codebase -- I think one of the integrations has a config which generates screenshots but I figured it would be ok to keep it out of the codebase for the future.

brekk:
@dguenther Thanks for the thorough feedback -- I've addressed everything and I think it should be cleaner now.

AmberKiso:
Looking forward to this being officially announced!

dguenther (approved):
Nice, thanks for fixing those up!

PR #222, opened 2022-04-06
Phase 2: User Details

brekk:
## Summary

1. Swapped out `MetricBox` for a newer, more generic component
2. Added two new event types -- `TRANSACTION_SENT` && `HOSTED_NODE`
3. Some trickery with tool tip info

This PR is still a *WORK IN PROGRESS* until we remove the very obvious hack and can replace it with real data.


components/user/EventRow.tsx
```diff
@@ -193,6 +225,14 @@ const summarizeEvent = (
         <Verbose>Promoted on {hostname}</Verbose>
       </>
     )
+  } else if (type === EventType.NODE_HOSTED) {
+    return false
```
dguenther, line 229: Maybe just return an empty react fragment for this if you don't want to deal with handling the boolean? I haven't looked at the implementation too closely.

PR #226, opened 2022-04-15
Update useForm `defaultValue` behaviors

brekk:
## Summary

1. Set the internal `touched` value based on a given `defaultValue`
2. Segment validation as a separate `useEffect`

## Testing Plan

This previously [merged PR](https://github.com/iron-fish/website-testnet/pull/225) was a quick fix. With this PR's changes, we can see the same behavior without the one-off `useEffect`.

1. Working: https://website-testnet-git-update-form-checkery-ironfish.vercel.app/login?email=cool@gmail.com
2. Also working (but invalid email): https://website-testnet-git-update-form-checkery-ironfish.vercel.app/login?email=coolcoolcoolemail

## Breaking Change

```
[ ] Yes
[x] Hopefully not
```

### DIFF
--- a/hooks/useForm.ts
+++ b/hooks/useForm.ts
@@ -76,8 +76,6 @@ export function useField(provided: ProvidedField): Field | null {
   const touched = defaultValue ? true : _touched
 
   const initiallyValid = defaultValue ? validation(defaultValue) : false
-  // eslint-disable-next-line no-console
-  console.log({ defaultValue, touched, _touched, initiallyValid })
   const [$value, $setter] = useState<string>(defaultValue)
   const radioOption =
     defaultRadioOption || (options && options[0] && options[0].value)
@@ -89,6 +87,8 @@ export function useField(provided: ProvidedField): Field | null {
   const [$disabled, $setDisabled] = useState<boolean>(false)
   const [, $setValid] = useState<boolean>(initiallyValid)
   const [$touched, $setTouched] = useState<boolean>(touched)
+  // eslint-disable-next-line no-console
+  console.log({ defaultValue, touched, _touched, $touched, initiallyValid })
   const [$field, $setField] = useState<Field | null>(null)
   const [$error, $setError] = useState<string>(defaultErrorText)
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


hooks/useForm.ts
```diff
@@ -83,13 +84,20 @@ export function useField(provided: ProvidedField): Field | null {
...
   const [$disabled, $setDisabled] = useState<boolean>(false)
-  const [, $setValid] = useState<boolean>(false)
+  const [$valid, $setValid] = useState<boolean>(initiallyValid)
   const [$touched, $setTouched] = useState<boolean>(touched)
   const [$field, $setField] = useState<Field | null>(null)
   const [$error, $setError] = useState<string>(defaultErrorText)
   useEffect(() => {
-    const valid = !$touched || ($touched && validation($value))
-    $setValid(valid)
+    if (defaultValue) {
```
danield9tqh, line 92: Is this necessary if we are already making the change to make `$valid` true on line 87? Seems like that would be enough to fix the validation error on submit for the original bug. But idk what `$touched` does. Does it influence some other logic?

dguenther: That's correct, the `if (defaultValue) {` portion of the effect isn't needed. It'll have the effect of never validating subsequent changes to `$value` unless `defaultValue` is unset.

dguenther (changes requested):

hooks/useForm.ts
brekk: I sliced out another `useEffect`, as simply removing the `defaultValue` clause didn't work (see [this commit](https://github.com/iron-fish/website-testnet/pull/226/commits/743d414298430af98b9c399534eac53d41128459) for what I tried). The core issue is whether `$touched` is set correctly -- if it isn't, the validation only happens after the user interacts with the field (which is the original issue Jason was seeing).

I'll investigate further (because this logic isn't ideal) but I'm inclined to largely keep things close to this commit as I think further refactoring will likely be a long process.

lwisne:
What is the status for this PR? Is it still to go live?

dguenther (approved):
Okay, I think this is better than a page-specific workaround, so let's get this in.

PR #227, opened 2022-04-18
Updating copy for the phase2 about page

leanthebean:
## Summary
Updating copy for the phase2 about page

## Testing Plan
Local

## Breaking Change

```
[X] No
[ ] Yes
```

### DIFF
--- a/components/About/data.tsx
+++ b/components/About/data.tsx
@@ -105,7 +105,7 @@ export const callsToAction = {
       points: ['12 hours of continuous uptime = 10 points'],
       status: Status.New,
       ctaText: 'Set up instructions',
-      href: '/faq',
+      href: '/faq#node-points',
     },
     {
       title: 'Submit a Pull Request',
@@ -136,7 +136,7 @@ export const callsToAction = {
       ),
       points: ['10 points for a transaction over 0.1 coins'],
       ctaText: 'Learn more',
-      href: '/faq',
+      href: '/faq#transaction-points',
       status: Status.New,
     },
     {
--- a/components/About/data.tsx
+++ b/components/About/data.tsx
@@ -105,7 +105,7 @@ export const callsToAction = {
       points: ['12 hours of continuous uptime = 10 points'],
       status: Status.New,
       ctaText: 'Set up instructions',
-      href: '[FAQ LINK]',
+      href: '/faq',
     },
     {
       title: 'Submit a Pull Request',
@@ -136,7 +136,7 @@ export const callsToAction = {
       ),
       points: ['10 points for a transaction over 0.1 coins'],
       ctaText: 'Learn more',
-      href: '[FAQ LINK]',
+      href: '/faq',
       status: Status.New,
     },
     {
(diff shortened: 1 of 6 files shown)


brekk (approved):

components/About/data.tsx
```diff
@@ -101,48 +101,43 @@ export const callsToAction = {
...
       content:
-        'Hosting a node using your graffiti automatically earns you points. You will only be rewarded points for hosting one node.',
-      points: ['12 hours = 100 points'],
+        'You can earn points by hosting an Iron Fish full node. To do so, make sure your telemetry is turned on, and your graffiti is set. You will only be rewarded points for hosting one node per graffiti.',
+      points: ['12 hours of continuous uptime = 100 points'],
       status: Status.New,
-      earn: 0,
-      disabled: true,
+      ctaText: 'Set up instructions',
+      href: 'https://coda.io/d/_dngSKECR9gO/Phase-2-FAQ_suBOW',
```
dguenther, line 108: This doc doesn't appear to be public yet -- is it ready to open up?

components/About/data.tsx
```diff
@@ -101,48 +101,43 @@ export const callsToAction = {
...
-      earn: 0,
       points: [
         'Small = 250 points',
         'Medium = 500 points',
         'Large = 1000 points',
       ],
       ctaText: 'Submit a PR',
-      // href: 'https://github.com/iron-fish/ironfish/pulls',
-      submissionForm: false,
+      href: 'https://github.com/iron-fish/ironfish/pulls',
```
dguenther, line 120: The text in this section mentions a form -- are we going to use the same form as last time? If so, we should switch this button to go to the form.

components/About/data.tsx
```diff
@@ -101,48 +101,43 @@ export const callsToAction = {
...
       title: `Sending a transaction`,
       content: (
         <>
-          Sending your mined $IRON earns you points. Use command{' '}
+          For every transaction you send to the Iron Bank of over 0.1 coins, you
+          get 10 points. To make this easier, we made a command that
+          automatically sends this transaction for you:
           <pre className="bg-iflightgray inline p-1 font-favorit">
-            send_transaction
+            ironfish deposit
```
dguenther, line 133: I believe it's going to be `testnet:deposit`, but we can readjust the docs when that work is ready.

components/About/data.tsx
```diff
@@ -101,48 +101,43 @@ export const callsToAction = {
...
-          . You get 1 point for each iron you send, as long as you send at least
-          1 iron.
         </>
       ),
-      points: ['1 $IRON sent = 1 point (minimum 1 $IRON)'],
-      ctaText: 'Get started with mining',
-      // href: 'https://ironfish.network/docs/onboarding/miner-iron-fish',
+      points: ['10 points for a transaction over 0.1 coins'],
+      ctaText: 'Learn more',
+      href: 'https://ironfish.network/docs/onboarding/miner-iron-fish',
```
dguenther, line 139: Button links to a doc page about mining, but the above text is about sending transactions using the `deposit` command -- maybe hiding the "Learn More" would be fine until we have something to put there?

dguenther (commented):
Commented on some things I noticed, I don't think any of these have to be blocking

components/About/data.tsx
leanthebean: yeah I'll change that

components/About/data.tsx
leanthebean: double checked with Mat that is indeed `ironfish deposit`

components/About/data.tsx
leanthebean: Updated

PR #245, opened 2022-05-04
Fix leaderboard link to phase 1

NullSoldier:
## Summary

The link to phase 1 goes to the wrong URL, because it doesn't have a
protocol so the browser treats it as a relative URL.

## Testing Plan

View the link on the leaderboard.

## Breaking Change

```
[ ] Yes
```


brekk (approved):
:shipit:

mat-if (approved):

PR #260, opened 2022-05-11
Change the leaderboard description slightly

dguenther:
## Summary

The leaderboard copy references net promoters, so I wanted to take that specific reference out of there. Open to other opinions on better copy.

Also this is targeted at master, not sure whether it'd be better to go into staging.

## Breaking Change

```
[ ] Yes
```


leanthebean (approved):

brekk (approved):
:shipit:

PR #261, opened 2022-05-11
About Page: pools have names and numbers

brekk:


utils/number-to-words.ts
```diff
@@ -0,0 +1,32 @@
+export const digit = (x: number): string => {
+  if (x === 9) {
```
lwisne, line 2: slightly cleaner here would be a k:v map

lwisne (approved):

PR #266, opened 2022-05-25
UI Fixes

wd021:
## Summary
Fixes for the following UI issues

**1 - Rank values are getting large, currently missing toLocaleString(). Also, need to truncate in the leaderboard row.**

Screenshot 2022-05-25 at 22 46 35 to localeString()
Screenshot 2022-05-25 at 22 46 44

**2- On the user page, when the graffiti is long, the container width is greater than the page width on small screens.**

Screenshot 2022-05-25 at 22 49 55

**3- On the user page, OffsetBorder Box can have height discrepancies in some cases. Also, reducing the font size to fit large numbers in box.**

Screenshot 2022-05-25 at 22 52 45

```
[ ] Yes
[x] No
```

### DIFF
--- a/utils/numberToOrdinal.ts
+++ b/utils/numberToOrdinal.ts
@@ -7,13 +7,13 @@ export function numberToOrdinal(num: number): string {
   const k = num % 100
 
   if (j === 1 && k !== 11) {
-    return `${num.toLocaleString()}st`
+    return `${num}st`
   }
   if (j == 2 && k != 12) {
-    return `${num.toLocaleString()}nd`
+    return `${num}nd`
   }
   if (j == 3 && k != 13) {
-    return `${num.toLocaleString()}rd`
+    return `${num}rd`
   }
-  return `${num.toLocaleString()}th`
+  return `${num}th`
 }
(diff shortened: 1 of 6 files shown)


utils/numberToOrdinal.ts
```diff
@@ -7,13 +7,13 @@ export function numberToOrdinal(num: number): string {
...
   if (j == 2 && k != 12) {
-    return `${num}nd`
+    return `${num.toLocaleString()}nd`
   }
   if (j == 3 && k != 13) {
-    return `${num}rd`
+    return `${num.toLocaleString()}rd`
   }
-  return `${num}th`
+  return `${num.toLocaleString()}th`
```
brekk, line 18: I could be wrong here but I don't think this is correct, largely because `toLocaleString` is locale / language specific, but our ordinal system function is always gonna return English-specific suffixes.

I don't think a formatted number like `8,400th` is correct, in the same way that in Germany the same number shouldn't be `8.400th`

brekk (commented):
Gonna ping our designer for review, code-wise it seems very reasonable except for the comment I made on the ordinal function.

utils/numberToOrdinal.ts
wd021: hmm, you might be right here. Instinctively, I felt like the 5 digit numbers needed a comma, even for the ordinal values. 

I'll revert that commit for now.

wd021:
still merging?

PR #286, opened 2022-09-26
Update LICENSE

pasha121:
## Breaking Change

```
[ ] Yes
```


dguenther:
This doesn't match the text of the MPL. For example, https://choosealicense.com/licenses/mpl-2.0/

PR #298, opened 2022-11-05
Fix deprecated description on About page

krompets:
On "About" page we have a deprecated description for "Rewards" section. The text is wrong because it does not include rewards for Phase 2. The right data are:
The Incentivized Testnet program will distribute up to 735,000 (1.75% of the initial supply) Iron Fish tokens to eligible participants, proportional to your Leaderboard points. Token distribution may be canceled at any time due to regulatory concerns. We may at any time amend or eliminate Leaderboard points.

Before:
image

After:
image

## Breaking Change

```
[ ] Yes
```


NullSoldier:
This isn't exactly valid, as it depends on the pools.

PR #307, opened 2022-11-10
Fix "Phase 1" points distribution text.

krompets:
## Summary

According to that "Prize pool 3" spans all phases there should be the same points depending on PR size.
But we see that in phase 1 Medium PR cost 750 points while in phase 2 it is 500 points.
Phase 1
image

Phase 2
image

We expect that PR cost should be the same for both phases. Also required changes in the backend. Because as I see the API returns 750 points for medium PR.

## Breaking Change

```
[ ] Yes
```


krompets:
Wrong branch. Also looks like we do not need update old site.

PR #313, opened 2022-11-14
chore: update header and remove countdowns

jowparks:
## Summary
Updates header 
and removes countdowns:

Screen Shot 2022-11-14 at 2 31 29 PM

## Breaking Change

```
[ ] Yes
```


pages/about.tsx
```diff
@@ -97,19 +96,15 @@ export default function About({ showNotification, loginContext }: AboutProps) {
           'flex-col'
         )}
       >
-        <CountdownTimer
-          end={new Date(1668542400000)}
-          event=" until end of Phase 2."
```
NullSoldier, line 102: inverse count down?

PR #321, opened 2022-11-16
chore: update background gradient

jowparks:
## Breaking Change

```
[ ] Yes
```


jowparks:
Abandoning since this would require a bunch of design rework due to gradient overlay, background buttons, etc.

PR #329, opened 2022-11-23
chore: floor countdown timer

jowparks:
## Summary
Do not render timer when it is complete
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/components/leaderboard/CountdownTimer.tsx
+++ b/components/leaderboard/CountdownTimer.tsx
@@ -24,7 +24,9 @@ const CountdownTimer = (props: CountdownTimerProps) => {
     return () => clearInterval(i)
   }, [])
   const ii = intervalToDuration({ start: $time, end })
-  return (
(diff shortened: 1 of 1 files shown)


components/leaderboard/CountdownTimer.tsx
```diff
@@ -24,7 +24,9 @@ const CountdownTimer = (props: CountdownTimerProps) => {
     return () => clearInterval(i)
   }, [])
   const ii = intervalToDuration({ start: $time, end })
-  return (
+  return now > end ? (
+    <></>
+  ) : (
     <div
       className="w-full flex flex-col justify-center items-center text-center"
       title={`Next week starts: ${formatEventDate(end)}`}
```
NullSoldier, line 32: Can we do what most websites do and just have it stuck at 0? Just add a floor?

NullSoldier (approved):

PR #272, opened 2022-06-28
Make opening outside links in the new tab.

krompets:
it is a good practice to open external links in a new tab. But currently, the links to social media in the footer are opening in the same tab. In this change added target="_blank" to make this small improvement.

## Breaking Change

Is this a breaking change? If yes, add notes below on why this is breaking and
what additional work is required, if any.

```
[ ] Yes
```

PR #351, opened 2023-01-07
Fixed crash while open user stats

krompets:
## Summary

We have an error when opening the user metrics page.
```
TypeError: Cannot read properties of null (reading 'toLocaleString')
    at De ([id]-21bb4c7fe40e7902.js:1:49840)
```
image
image

This change is a fix for this problem.

AFTER FIX:
image

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/pages/users/[id].tsx
+++ b/pages/users/[id].tsx
@@ -165,7 +165,7 @@ export default function User({ showNotification, loginContext }: Props) {
     ? numberToOrdinal($allTimeMetrics.pools.main.rank)
     : 'NA'
 
-  const phase3Points = $allTimeMetrics.pools.main.points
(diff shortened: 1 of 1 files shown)


NullSoldier:
@krompets What graffiti do you want this credited to?

krompets:
@NullSoldier  this one was credited (graffiti: testnet-ironfish21). {"id": 85935556, "user_id": 7077,"points": 250}
I found two PRs without credits (form sent). But not sure if they should be credited:
https://github.com/iron-fish/wallet-app/pull/186
https://github.com/iron-fish/website-testnet/pull/272

NullSoldier:
We're processing the backlog now.

PR #357, opened 2023-01-16
Remove deplicated category pull-requests in options

hairtail:
## Summary
We use Pool3 for pull request points, so this "pull requests" category is duplicated, remove it.
## Breaking Change

```
[ ] Yes
```


NullSoldier:
That's not clear that Pool 3 is pull requests. Someone who doesn't know that may look in the list and wonder why they can't filter by pull requests.

PR #371, opened 2023-01-19
fix: add faqs from recent launch

jowparks:
## Summary
SIGSEGV and where are my points, two most commonly asked questions.
## Testing Plan

Screen Shot 2023-01-19 at 11 15 24 AM

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/pages/faq.tsx
+++ b/pages/faq.tsx
@@ -75,8 +75,9 @@ const questions: ReadonlyArray<{
     answer: (
       <>
         You can mint an asset by running the command below. Make sure you use
-        your graffiti as your asset name.
-        <div>ironfish wallet:mint</div>
+        your graffiti as your asset name:
+        <br />
+        <code>ironfish wallet:mint</code>
       </>
     ),
   },
@@ -85,8 +86,9 @@ const questions: ReadonlyArray<{
     id: `how-to-burn`,
     answer: (
       <div>
-        You can burn an asset by running the command{' '}
-        <div>ironfish wallet:burn</div>
+        You can burn an asset by running the command:
+        <br />
+        <code>ironfish wallet:burn</code>
       </div>
     ),
   },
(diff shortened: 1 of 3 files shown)


NullSoldier:
The sigsev crash is not going to really help anyone, particularly because it's just a known issue and not a FAQ question with a useful answer.

I don't think the points question is going to help. 99% of issues people have is that they are just running the incorrect commands we posted. Will they see this and use it to fix their issues? Maybe but I think we need to simplify it.

pages/faq.tsx
```diff
@@ -219,7 +219,48 @@ const questions: ReadonlyArray<{
     question:
       'I get the error, "Your database needs to be upgraded (v10 vs v14)."',
     id: 'database-error',
-    answer: `You are running phase 3 on an old phase 2 database. You need to delete your datadir located at ~/.ironfish`,
+    answer: (
+      <div>
+        You are running phase 3 on an old phase 2 database. You need to delete
+        your old datadir. Please first try the command{' '}
+        <code>ironfish reset</code>. If that does not work try{' '}
```
NullSoldier, line 226: Until we release we should not recommend ironfish reset, then we should only recommend it

pages/faq.tsx
```diff
@@ -219,7 +219,48 @@ const questions: ReadonlyArray<{
...
+        We are aware of this issue, please submit your crash report{' '}
+        <span className="underline">
+          <Link href="https://github.com/iron-fish/ironfish/issues/2985">
+            here
+          </Link>
+        </span>{' '}
+        and follow along for more details.
+      </div>
+    ),
+  },
```
NullSoldier, line 245: I don't think anyone is gonna look at the FAQ for this, it's gonna be better in a "known issues list". I also don't like to put temporary things into the FAQ because as soon as we fix this we need to come in here and remove this.

pages/faq.tsx
```diff
@@ -221,7 +221,48 @@ const questions: ReadonlyArray<{
...
+      <div>
+        Make sure you have followed the debugging instructions on{' '}
+        <span className="underline">
+          <Link href="https://github.com/iron-fish/ironfish/wiki/Phase-3-Debugging">
+            the wiki.
+          </Link>
+        </span>{' '}
+        If that does not solve your problem, come post in the{' '}
+        <span className="underline">
+          <Link href="https://discord.gg/ironfish">Discord server</Link>
```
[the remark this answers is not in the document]
jowparks: Got that from existing link above in the page, will update there too.

jowparks:
No longer needed

PR #374, opened 2023-01-19
Fix link for events from phase3

krompets:
## Summary
We have an issue with links to explorer for MINT|BURN|SENT events.
Current result:
We cannot navigate to explorer. We only can copy transaction hash.

After fix:
The links are clickable and we can navigate to the explorer
image
 And after clicking we see our transaction in explorer
image

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier:
@krompets What graffiti is this for?

PR #375, opened 2023-01-19
Changed SNAKE_CASE text to regular like it done for other categories

krompets:
## Summary
In the "activity" table we have snake_case for new activities while based on previous phases looks like it should be regular text

### AFTER FIX:
image

### BEFORE FIX:
image

## Breaking Change

```
[ ] Yes
```


NullSoldier (approved):

NullSoldier (approved):

NullSoldier:
@krompets What graffiti is this for?

PR #380, opened 2023-01-23
Fixed wrong weeks interval

krompets:
## Summary

The week's interval is wrong. Due to this issue, many participants were confused because in the table we saw points for Send, Burn, and Mint assets twice while the weekly limit is 1.
See:
image

After fix:
image

## Breaking Change

```
[ ] Yes
```


krompets:
Already done in https://github.com/iron-fish/website-testnet/pull/381

PR #385, opened 2023-02-03
feat: Add Google Recaptcha v3 to sign-up form

dgca:
## Summary

Adds Google Recaptcha v3 to the sign up form. Passes the recaptcha token as a `recaptcha` field to the `/users` POST endpoint.

I also added the `NEXT_PUBLIC_RECAPTCHA_KEY` variable to all environment on `website-testnet` on Vercel.

## Testing Plan

Manual testing

## Breaking Change

```
[ ] Yes
```


apiClient/client.ts
```diff
@@ -20,20 +20,32 @@ import {
...
+
+export async function createUser({
+  email,
+  graffiti,
+  socialChoice,
+  social,
+  country: country_code,
+  recaptcha,
+  github,
+}: Params): Promise<ApiUser | ApiError> {
```
dgca, line 41: Did this so we don't have to rely on parameter order to build the right payload.

dguenther:
I didn't hit the recaptcha for this when I created an account, but maybe it has something to do with the domain + sending the environment variable on staging?

dguenther (approved):
Hitting approve on this, I don't think it's absolutely necessary to test in staging before rolling it out. I'd definitely like to test it on prod though

NullSoldier (approved):

PR #395, opened 2023-02-27
Remove unnecessary countdown text

wd021:
time until next block text no longer needed.


wd021:
looks like this has been resolved. closing.

PR #406, opened 2023-03-03
add a warning box about exporting wallet key

lwisne:
## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/components/Airdrop/StepSubmitAddress/StepSubmitAddress.tsx
+++ b/components/Airdrop/StepSubmitAddress/StepSubmitAddress.tsx
@@ -4,19 +4,20 @@ import Button from 'components/Button'
 import TextField from 'components/Form/TextField'
 import { UNSET } from 'utils/forms'
 import { useField, WHITESPACE } from 'hooks/useForm'
+import { useState } from 'react'
 
 const publicAddressField = {
   id: 'address',
   label: 'Public Address',
   placeholder: 'Your Iron Fish public address',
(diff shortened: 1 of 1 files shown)


lwisne:
There is some "Lawrence does HTML" janky stuff here, feel free to tell me the right way to do it.

Specifically, I couldn't easily figure out how to do the CSS for having the "Next" button appear disabled. I assume there's some magic class.

components/Airdrop/StepSubmitAddress/StepSubmitAddress.tsx
```diff
@@ -52,18 +58,51 @@ export default function StepSubmitAddress({ onNext }: Props) {
...
+            </a>{' '}
+            to ensure you always have access.
+          </div>
+          <div style={{ marginTop: '1em' }}>
+            <input
+              type="checkbox"
+              onClick={() => {
+                setButtonDisabled(!buttonDisabled)
+              }}
+            ></input>{' '}
```
dgca, line 88: nit: This works, but it's indirect since we're relying on the fact that a checkbox is unchecked by default.

A more declarative way to do would be:

```js
const [isChecked, setIsChecked] = useState(false);
...
<input
  type="checkbox"
  checked={isChecked}
  onChange={e => {
    setIsChecked(e.target.checked)
  }}
/>
...
<button disabled={!isChecked}>hi</button>
```

Feel free to update it or keep it as is.

PR #434, opened 2023-03-17
Remove sign up page to prevent new users sign up

ygao76:
## Summary
Show error message when getUserDetails api failed
Screen Shot 2023-03-17 at 4 45 09 PM

Disable sign up page
Screen Shot 2023-03-17 at 4 44 43 PM

Remove sign up link on log in page
Screen Shot 2023-03-17 at 4 46 18 PM

## Breaking Change

```
[ ] Yes
```


pages/login.tsx
```diff
@@ -42,7 +41,7 @@ type LoginProps = {
 export default function Login({ showNotification, loginContext }: LoginProps) {
   const $router = useRouter()
   const { status } = useProtectedRoute({
-    ifLoggedIn: `/leaderboard?toast=${btoa("You're already logged in.")}`,
+    ifLoggedIn: `/dashboard?toast=${btoa("You're already logged in.")}`,
```
danield9tqh, line 44: If we are removing login page why is it sending to `/dashboard` instead of `/leaderboard`? Isn't dashboard only for if login is enabled?

pages/signup.tsx
```diff
@@ -111,100 +103,100 @@ export default function SignUp({
...
-  const testInvalid = useCallback(() => {
-    const noEmail = !$email?.touched
-    const noGraffiti = !$graffiti?.touched
-    const noGithub = !$github?.touched
-    // for old men
-    const noCountry = !$country?.touched
-    const untouched = noEmail || noGraffiti || noCountry
-    const invalid =
-      !$email?.valid || !$graffiti?.valid || !$social?.valid || !$country?.valid
+  // const testInvalid = useCallback(() => {
```
danield9tqh, line 116: Is there a better way we could remove this functionality other than just commenting it all out? Looks pretty messy + there's not reasoning for when future people come to look at the code

pages/login.tsx
ygao76: We are removing sign up page

PR #440, opened 2023-03-21
Render failure reason from API

NullSoldier:
## Summary

This now renders the failure reason including the help URL according to figma.

Screen Shot 2023-03-20 at 11 18 17 PM

## Breaking Change

```
[ ] Yes
```

### DIFF
--- a/components/Airdrop/hooks/useApprovalStatusChip.tsx
+++ b/components/Airdrop/hooks/useApprovalStatusChip.tsx
@@ -5,28 +5,49 @@ import { getNextEligiblePhase } from './usePhaseStatus'
 import type { KycConfig, KycStatus } from '../types/JumioTypes'
 import { titlesByPhase } from '../RewardItem/RewardItem'
 
-function withDetails(message: string, details?: string) {
+function withDetails(
+  message: string,
+  details?: string,
+  helpUrl?: string
+): JSX.Element {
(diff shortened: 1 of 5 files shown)


components/Airdrop/hooks/useApprovalStatusChip.tsx
```diff
@@ -5,28 +5,49 @@ import { getNextEligiblePhase } from './usePhaseStatus'
...
 
-  return `${message}: ${details}`
+  return (
+    <span>
+      {message}: {details}
+      {helpUrl ? (
+        <>
+          {' '}
+          <a href={helpUrl} target="_blank" rel="noreferrer">
+            Learn more here
```
jowparks, line 24: underline

jowparks (approved):
Small nit, looks good, thanks for screenshot

skylarrichard123 (changes requested):
links should have underline. No need to say "KYC failed: YOU FAILED". I think the user would understand that from "KYC failed".

components/Airdrop/hooks/useApprovalStatusChip.tsx
NullSoldier: fixed

NullSoldier:
The failed is a standin for the real message because the API is not sending it back yet.

PR #447, opened 2023-04-05
Check jwt token in loginContext

ygao76:
## Summary

The new login flow set jwt token in the cookie, example of cookie string
_ga=GA1.1.1480516078.1663010890; _ga_GJD73W9V3M=GS1.1.1680053950.84.0.1680053950.0.0.0; ironfish_jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ5YWp1bi5nYW8xMDA0QGdtYWlsLmNvbSIsImlhdCI6MTY4MDY0OTM5OTU2NywiZXhwIjoxNjgwNjQ5NDg1OTY3fQ.QCRG_2Q30bSYQXAmpAEuPx8XqaNFQ8YbVxBtDWKcyMA

When access the /dashboard with valid jwt token, we should let the user through. This PR adds the logic to check jwt token in LoginContext. It will call backend /me endpoint to validate the token. If /me response has 200 status code, the user pass the auth.

## Testing Plan
Vercel

## Breaking Change

```
[ ] Yes
```
@dgca @rohanjadvani

### DIFF
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -235,21 +235,17 @@ export async function getUserDetails(
   token?: string
 ): Promise<ApiUserMetadata | ApiError | LocalError> {
   try {
-    let response
-    if (!token) {
-      response = await fetch(`${API_URL}/me`, {
-        headers: {
-          'Content-Type': 'application/json',
-        },
-      })
-    } else {
-      response = await fetch(`${API_URL}/me`, {
-        headers: {
-          'Content-Type': 'application/json',
-          Authorization: `Bearer ${token}`,
-        },
-      })
+    const headers: HeadersInit = {
+      'Content-Type': 'application/json',
     }
+
+    if (token) {
+      headers.Authorization = `Bearer ${token}`
+    }
+
+    const response = await fetch(`${API_URL}/me`, {
+      headers,
+    })
 
     const data = await response.json()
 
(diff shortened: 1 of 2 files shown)


hooks/useLogin.ts
```diff
@@ -116,6 +131,20 @@ export function useLogin(config: LoginProps = {}) {
...
+  for (let i = 0; i < ca.length; i++) {
+    const c = ca[i].trim()
+
+    if (c.indexOf(name) == 0) {
+      return c.substring(name.length, c.length)
+    }
+  }
+
+  return ''
+}
```
dgca, line 146: If you want to avoid loops and go with something a little more declarative:

```ts
function getCookie(name: string) {
  return document.cookie
    .split(/;\s?/)
    .find((item) => item.startsWith(`${name}=`))
    ?.slice(name.length + 1) ?? '';
}
```

apiClient/client.ts
```diff
@@ -232,15 +232,24 @@ export async function login(email: string): Promise<any> {
...
+        },
+      })
+    } else {
+      response = await fetch(`${API_URL}/me`, {
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${token}`,
+        },
+      })
+    }
```
dgca, line 252: Suggestion: You can do this to avoid duplicating the fetch call:

```ts
  const headers: HeadersInit = {
    'Content-Type': 'application/json',
  }

  if (token) {
    headers.Authorization = `Bearer ${token}`
  }

  const response = await fetch(`${API_URL}/me`, {
    headers,
  })
```

dgca (approved):

PR #451, opened 2023-04-07
Call server to logout

ygao76:
## Summary
Call /logout endpoint to empty jwt token in cookie

## Breaking Change

```
[ ] Yes
```
@dgca @rohanjadvani

### DIFF
--- a/apiClient/client.ts
+++ b/apiClient/client.ts
@@ -231,6 +231,20 @@ export async function login(email: string): Promise<any> {
   }
 }
 
+export async function logout(): Promise<boolean> {
+  try {
+    const response = await fetch(`${API_URL}/logout`)
+
+    if (response.status !== 200) {
+      return false
+    }
+
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
 export async function getUserDetails(
   token?: string
 ): Promise<ApiUserMetadata | ApiError | LocalError> {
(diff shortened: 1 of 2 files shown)


apiClient/client.ts
```diff
@@ -231,6 +231,20 @@ export async function login(email: string): Promise<any> {
   }
 }
 
+export async function logout(): Promise<boolean> {
+  try {
+    const response = await fetch(`${API_URL}/logout`)
```
rohanjadvani, line 236: Should this be a post?

pages/logout.tsx
```diff
@@ -14,7 +14,10 @@ export default function Logout({ loginContext }: LogoutProps) {
   useEffect(() => {
     const sayGoodbye = async () => {
       const loggedOut = await magic?.user.logout()
-      if (loggedOut) {
+
+      const jwtLoggedOut = await logout()
```
rohanjadvani, line 18: Should we only call this if the magic user doesn't exist?

pages/logout.tsx
```diff
@@ -13,7 +13,12 @@ type LogoutProps = {
 export default function Logout({ loginContext }: LogoutProps) {
   useEffect(() => {
     const sayGoodbye = async () => {
-      const loggedOut = await magic?.user.logout()
+      let loggedOut = await magic?.user.logout()
+
+      if (loggedOut === undefined || loggedOut == false) {
```
rohanjadvani, line 18:
```suggestion
      if (!loggedOut) {
```

rohanjadvani (approved):

